PR #56, opened 2025-03-17
Reduce the number of targets evaluations by skipping the first couple of targets known to be invalid

ricohageman:


ricohageman:
Noticed https://github.com/tscircuit/capacity-autorouter/pull/55/commits/e68150953db5607619466d27966354137db6ae45 which has a much better improvement

PR #57, opened 2025-03-17
Speedup SingleHighDensityRouteSolver using priorityQueue

ricohageman:
Reduce the time spent sorting candidates by keeping them sorted inside a priority queue. This yields a nice 17% speedup on the ledmatrix3 benchmark (from 4.74s to 3.88s after including the changes in https://github.com/tscircuit/capacity-autorouter/pull/55)


package.json
```diff
@@ -36,7 +36,8 @@
     "tsup": "^8.3.6",
     "use-mouse-matrix-transform": "^1.3.0",
     "vite": "^6.0.11",
-    "vite-tsconfig-paths": "^5.1.4"
+    "vite-tsconfig-paths": "^5.1.4",
+    "@datastructures-js/priority-queue": "^6.3.2"
```
ricohageman, line 40: This seems to be an often used package for priority queues, but let me know if you prefer to use a local variant of the datastructure to reduce the dependencies.

lib/solvers/HighDensitySolver/SingleHighDensityRouteSolver.ts
```diff
@@ -411,22 +411,24 @@ export class SingleHighDensityRouteSolver extends BaseSolver {
...
 
-    if (!currentNode) {
+    if (!currentNode || !currentNodeKey) {
       this.failed = true
       return
     }
-    this.exploredNodes.add(this.getNodeKey(currentNode))
-    this.debug_exploredNodesOrdered.push(this.getNodeKey(currentNode))
+    this.exploredNodes.add(currentNodeKey)
+    this.debug_exploredNodesOrdered.push(currentNodeKey)
```
ricohageman, line 431: Small micro optimisation to reduce the number of times the key is generated

seveibar:
looks pretty good. Yea I generally don't want to have external dependencies because they often aren't super well optimized (even the popular ones). Can you try re-implementing the priority queue? it's a relatively simple data structure. I would also stay away from "making it generic", we can have a domain-specific priority queue (e.g. `SingleRouteCandidatePriorityQueue`) because callbacks or custom compare fns etc. often have a small overhead.

ricohageman:
> Wish we had a way to measure the perf of these improvements easily, maybe we can create something like we did for tscircuit/core soon that uses the historical value from localstorage to create a graph for each run
> 
> image

Maybe something like codspeed https://codspeed.io/ could help here as well

lib/solvers/HighDensitySolver/SingleHighDensityRouteSolver.ts
seveibar: niceeee

seveibar (approved):
17% is really solid, nice work!

The license on this repo was committed incorrectly before. Just wanted to double-check you saw that license change prior to merging (this repo is MIT, this is the pattern for all tscircuit repos that aren't forks/required to use another license)

seveibar:
Yea that codspeed service looks pretty good! bookmarked it and will be taking a look soon

ricohageman:
> The license on this repo was committed incorrectly before. Just wanted to double-check you saw that license change (from empty file to MIT) prior to merging (this repo is MIT, this is the pattern for all tscircuit repos that aren't forks/required to use another license)

Yeap with the MIT license introduced in 
560a144c9920bb4dcce7ebfcb963b30939b65e91

ricohageman:
Benchmarked once more with the current main and the relative performance remains similar (2.5s -> 2.0s -20%)

seveibar:
thanks again @ricohageman 😁

PR #107, opened 2025-04-26
Speedup Pathing solver by re-using the nodeEdgeMap

ricohageman:
A lot of time in the pathing solver was spent creating the nodeEdgeMap. This is wasting time as the edges do not change. By re-using the nodeEdgeMap the runtime of keyboard5 reduces from 8.39s to 3.72s (-55%).

Before:
Screenshot 2025-04-26 at 22 45 43

After:
Screenshot 2025-04-26 at 22 41 19


seveibar:
holy moly how did I not see this. NICE @ricohageman !!!

seveibar:
(i will fix the merge conflict shortly! i was doing a bunch of development and would have loved to have this for my own velocity haha)

seveibar:
@ricohageman i wasn't able to see the same performance boost, I'm wondering if this is because of some other change or maybe I resolved the conflict improperly (note: I used a rebase and force push). I think I did it correctly, but I'm only seeing a small speedup

I THINK what is happening (and I could be wrong) is the hyper parameters have changed to diminish the effect of this change. I increase the size of sections and decreased the attempts per node using some insights from the "Stats" box for the `pathingOptimizer` (previously was implemented as the second part of the `pathingSolver`)

This is still a great contrib and will be useful. You may have noticed some performance degradation recently, this is because I'm attempting to move towards a cache-optimized solution. Most of the phases that take a lot of time (any of the section solvers) are theoretically super cache-able. In addition to speculative performance benefits from the cache, the section solvers tend to be good at making sure there are no "stupid, easily fixed mistakes" since they examine nodes with high probabilities of failure.

The specific change that reduced the effectiveness of this fix (I think) is the HyperCapacityPathingSingleSectionSolver, which runs Hyper/Meta-A* for 10 solvers with different path orderings. We need less MAX_ATTEMPTS_PER_NODE because of the hyper solver, so it was reduced from 10 to 1.

image


image

image

ricohageman:
> ... decreased the attempts per node (from 10 to 1) ..

@seveibar I suspect that's indeed the reason. Just checked and I think you've resolved it properly, it also doesn't show up in the performance profile as often as it did before the fix.

PR #174, opened 2025-06-26
Add fixture for cn1936, cn2576 HyperHighDensityDebugger

techmannih:


seveibar (changes requested):
highdensity76 is working

seveibar:
image

PR #180, opened 2025-06-30
Fix caching issue by rounding the deltas with a higher precision

ricohageman:
### Why are we changing this?
We've observed cases where the `unravelMultiSectionSolver` was not completely caches during the second time it was solved. The working theory was that there were some rounding issues causing this.

### What has changed?
Delta values used in in the cached values of the `unravelMultiSectionSolver` are now rounded to a higher precision than the coordinates. This prevents issues from accumulated errors due to rounding.

### Visualisation
https://streamable.com/yr0hm1

### DIFF
--- a/lib/solvers/UnravelSolver/CachedUnravelSectionSolver.ts
+++ b/lib/solvers/UnravelSolver/CachedUnravelSectionSolver.ts
@@ -37,6 +37,13 @@ const approximateCoordinate = (coord: number): string => {
   return (Math.round(coord * 20) / 20).toFixed(2)
 }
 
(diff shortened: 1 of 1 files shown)


ricohageman:
@seveibar the unravel solver is still not instant for larger boards (e.g. keyboards). But I think that's caused by the fact that the stopping condition for the unravel solver is rather rough. It only ends after a certain number of iterations without improvements or when there are no more candidates to explore. For large boards it makes sense that there are always candidates to explore, so it will just try a lot before it stops due to the iterations limit. Might be something that we could add to the cache?

seveibar (approved):
WOAH nice fix!!!!!

PR #181, opened 2025-07-01
Prevent cache key collisions when multiple routes exist between the same node pair

ricohageman:
### Why are we changing this?
Even after merging https://github.com/tscircuit/tscircuit-autorouter/pull/180 we still observed issues with the cache where a lot of easily cacheable actions were missed. When the cache of the `HyperCapacityPathingSolver` was disabled, the `UnravelSolver` would reach a 100% cache rate. In addition, the logs showed a lot of `Could not find cache space connection ID` warnings. This indicates an issues with the way the `HyperCapacityPathingSolver` performs caching. Claude analysed the code and highlighted that collisions due to multiple paths between two points were the issue.

### What has changed?
- Adhere to the iteration limit provided as the input of the `HyperCapacityPathingSolver`
- Include statistics of the `HyperCapacityPathingSolver` cache in the output
- Create an unique identifier for each path between two points

### Visualisation
https://streamable.com/fnqae6


seveibar (approved):
I think this is a step in the right direction for demonstrating the cache key content issue, technically the new extended cache key shouldn't be necessary, but it's a nice demonstration of the underlying issue and provides a short-term fix

Issue #178, opened 2025-06-30
Autorouting Bug Report: <group#602 name=".LedN" />


Anshgrover23:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=2807d7a9-517f-4824-831f-ff6b2d2d299a

(any additional details? Maybe a screenshot?)

PR #192, opened 2025-07-07
Enable closed form high density routing again

ricohageman:
### What has changed?
- Added fixture of https://github.com/tscircuit/tscircuit-autorouter/issues/178
- Enabled closed form high density routing again

### Visualisation
Screenshot 2025-07-07 at 23 01 52

https://capacity-node-autorouter-git-closed-form-crossing-tscircuit.vercel.app/renderer.html?fixtureId=%7B%22path%22%3A%22examples%2Fbug-reports%2Fbug-report.fixture.tsx%22%7D&locked=true&bug_report_id=2807d7a9-517f-4824-831f-ff6b2d2d299a

Fixes https://github.com/tscircuit/tscircuit-autorouter/issues/178

### DIFF
--- a/examples/assets/cn7890-nodeWithPortPoints.json
+++ b/examples/assets/cn7890-nodeWithPortPoints.json
@@ -1,0 +1,205 @@
+{
+  "nodeId": "cn7890",
+  "capacityMeshNode": {
+    "capacityMeshNodeId": "cn7890",
+    "center": {
+      "x": 66.796875,
+      "y": -18.359375
+    },
+    "width": 0.78125,
+    "height": 0.78125,
+    "layer": "top",
+    "availableZ": [
+      0,
+      1
+    ],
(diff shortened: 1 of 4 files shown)


seveibar:
I'm checking out the preview, i think it may cause regressions but unfortunately we don't seem to have a test that captures any regressions

seveibar (approved):
well until it's obvious where this caused a regression, i think we can just add it back in and next time we disable it (if we do) we should create a test, I wish i had said more in the PR that disabled it https://github.com/tscircuit/tscircuit-autorouter/pull/148

ricohageman:
> one difference between this PR and the PR that disabled it is it's now listed as the last solver- maybe it's not being called as early now and is therefore used less 🤷

Yeah I've moved it down based on your comment. Basically, if the other algorithms could not solve it, use the closed form solution

Issue #191, opened 2025-07-07
Multi Head Solver only running for one iteration (highdensity79)


seveibar:
https://github.com/tscircuit/tscircuit-autorouter/pull/188

PR #193, opened 2025-07-07
Improved handling of the `ViaPossibilitiesSolver` part of the `MultiHeadSolver`

ricohageman:
### Why are we changing this?
We've observed that the MultiHeadSolver failed after a single iteration. This was caused by the `ViaPossibilitiesSolver2` not finding a solution within the iteration limit. This pull request introduces a couple of fixes for the same problem.

### What has changed?
- The entire MultiHeadSolver no longer fails directly when a single iteration of the `ViaPossibilitiesSolver2` doesn't find a solution
- The `ViaPossibilitiesSolver2` could not find a solution because it iteratively added a via just slightly closer to the target. By adding a floating point tolerance this is no longer the case
- Limit the search to a maximum number of vias for a single trace within a node

Fixes https://github.com/tscircuit/tscircuit-autorouter/issues/191

### Visualisation
[Highdensity fixture 79](https://capacity-node-autorouter-git-via-possibilities-tscircuit.vercel.app/?fixtureId=%7B%22path%22%3A%22examples%2Fhighdensity%2Fhighdensity79.fixture.tsx%22%7D) can now be solved while this was not previously the case
Screenshot 2025-07-08 at 00 09 35

### DIFF
--- a/lib/solvers/HighDensitySolver/MultiHeadPolyLineIntraNodeSolver/MultiHeadPolyLineIntraNodeSolver3_ViaPossibilitiesSolverIntegration.ts
+++ b/lib/solvers/HighDensitySolver/MultiHeadPolyLineIntraNodeSolver/MultiHeadPolyLineIntraNodeSolver3_ViaPossibilitiesSolverIntegration.ts
@@ -48,9 +48,6 @@ export class MultiHeadPolyLineIntraNodeSolver3 extends MultiHeadPolyLineIntraNod
     viaSolver.solve()
 
     if (viaSolver.failed || !viaSolver.solved) {
-      console.error(
-        viaSolver.error ?? "ViaPossibilitiesSolver2 failed to find a solution.",
-      )
       return null
     }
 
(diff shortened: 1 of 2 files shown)


lib/solvers/HighDensitySolver/MultiHeadPolyLineIntraNodeSolver/MultiHeadPolyLineIntraNodeSolver3_ViaPossibilitiesSolverIntegration.ts
```diff
@@ -48,10 +48,9 @@ export class MultiHeadPolyLineIntraNodeSolver3 extends MultiHeadPolyLineIntraNod
     viaSolver.solve()
 
     if (viaSolver.failed || !viaSolver.solved) {
-      this.failed = true
-      this.error =
-        viaSolver.error ?? "ViaPossibilitiesSolver2 failed to find a solution."
-      console.error(this.error)
+      console.error(
+        viaSolver.error ?? "ViaPossibilitiesSolver2 failed to find a solution.",
+      )
```
ricohageman, line 53: We might be able to remove this logging completely, it's not really relevant to know that one iteration of the `ViaPossibilitiesSolver2` failed

seveibar: yes i agree, it goes off a lot in console as well!! We should probably have all of our logging done within a ctx.logger so that it can be enabled/disabled depending on environment

ricohageman:
> interesting- is this because the points are overlapping 🤔

That's what I expect indeed. We have 3 points of which 2 are of the form `(x, y, 1) -> (x2, y2, 0)` & `(x, y, 0) -> (x2, y2, 1)`. So they need to go the same place and have a via. Their midpoints obviously overlap, and every step towards the intersection is very small, causing the iterations to exceed their limit

PR #197, opened 2025-07-12
added highdensity fixture for node cn541

ShiboSoftwareDev:


seveibar:
What is the source bug report this is from @ShiboSoftwareDev ? We're trying to figure out how these near-impossible high density problems are getting created

ShiboSoftwareDev:
@ricohageman the reproduction: comment out the connection to J3: https://tscircuit.com/editor?package_id=514001a7-a8dd-4e7a-8824-9d68ba57f42e
image

ShiboSoftwareDev:
nevermind, here is the bug report: https://github.com/tscircuit/tscircuit-autorouter/issues/200

ricohageman:
@ShiboSoftwareDev that would be great! Otherwise it's hard to evaluate the problem in the wider context and determine if another step of the algorithm is actually the culprit

PR #219, opened 2025-07-25
added highdensity85 autorouter bug repro

Abse2001:


seveibar:
this is a really great repro

PR #248, opened 2025-09-20
Fix #247, Don't remove straw nodes that are dead ends, create edges between straws with same parent capacity mesh node

### LINKED ISSUE #247: Autorouting Bug Report: <board#168 />
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=70a81dc2-2c28-4380-8d57-849a3c32198c

(any additional details? Maybe a screenshot?)
routing fails.

Heinrich-XIAO:
Fixes #247  by connecting straws in edge solver.
/claim #247


Heinrich-XIAO:
don't mind the commits before fb72737. those are from a previous branch that i branched off from.

seveibar:
@Heinrich-XIAO i've examined an increase the the time for the keyboard solver from 11.3s to ~14s (+20% duration)

image


I understand what's happening now but this is probably not the right solution. Straws nodes should not travel to eachother. Video to follow

PR #317, opened 2025-11-15
implement Z-separation real-time slider control

0hmX:


seveibar:
I don't mind the UI- but i feel like it'd be better implemented by adding `z` to rects inside `graphics-debug` and trying to implement it natively in graphics debug, possibly allowing "fading" and disabling mouse events on other unselected z layers, and reordering to the "selected z layer"

PR #324, opened 2025-11-17
Prevents autorouter failures when microcontrollers have internally connected power pins

0hmX:
close: https://discordapp.com/channels/1233487248129921135/1439406857537192078

Before:
image

After:
image

Adds internallyConnectedPointIds to prevent routing failures when microcontrollers have internally connected pins (like RP2040 IOVDD). The NetToPointPairsSolver now treats internal groups as single virtual points, routing only one representative to external components.

### DIFF
--- a/tests/examples/internal-connections-only.test.ts
+++ b/tests/examples/internal-connections-only.test.ts
@@ -8,10 +8,7 @@ import "tests/fixtures/svg-matcher"
 
 test("internal-connections-only example", async () => {
   const internalConnectionsOnlySrj: SimpleRouteJson = JSON.parse(
-    fs.readFileSync(
-      "assets/internal-connections-only.srj.json",
-      "utf-8",
-    ),
+    fs.readFileSync("assets/internal-connections-only.srj.json", "utf-8"),
   )
   const solver = new AutoroutingPipelineSolver(internalConnectionsOnlySrj)
   solver.solve()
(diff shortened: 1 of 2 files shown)


imrishabh18 (commented):
Nice! I think after this we need to pass the `internallyConnectedPins` to the simpleRouteJson in core

lib/types/srj-types.ts
```diff
@@ -35,6 +35,7 @@ export interface SimpleRouteConnection {
     pcb_port_id?: string
   }>
   externallyConnectedPointIds?: string[][]
```
seveibar:
```suggestion
```

lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
```diff
@@ -61,24 +80,58 @@ export class NetToPointPairsSolver extends BaseSolver {
...
+        // No routing required – they are already connected
         return
       }
       this.newConnections.push(connection)
       return
     }
 
-    const edges = buildMinimumSpanningTree(connection.pointsToConnect)
+    // Create a representative point for each internal group
+    const internalGroupRepresentatives = new Map<
```
seveibar: not clear at all

seveibar:
@0hmX you've got to start over rather than piling on, _start_ with a simple visual snapshot

seveibar:
@0hmX i believe this would be the simplest possible example:

image

PR #327, opened 2025-11-19
feat: Implement off-board connection substitution

0hmX:
Implements a new routing optimization feature that uses  connections to find more efficient paths.

When a connection is marked , its points are treated as electrically equivalent. The  now preprocesses these connections to create equivalence groups. When solving for an on-board connection, the solver can now substitute a target point with a closer, equivalent point from an off-board group, reducing the total trace length.


tests/offboard-substitution/transitivity.test.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * This creates a single equivalence group of (A, B, C). The solver must
+ * evaluate the path from X to every point in this group.
+ *
+ * Expected Outcome:
+ * The solver should identify that the path from X(0,0) to C(1,1) is the
+ * shortest possible route and create a trace between them, ignoring the
+ * original A(10,10) and B(20,20) targets.
+ */
+test("transitivity", () => {
```
seveibar:
```suggestion
test("connection.isOffBoard transitivity", () => {
```

lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
```diff
@@ -20,14 +25,79 @@ import { seededRandom } from "lib/utils/cloneAndShuffleArray"
...
+          Math.pow(p1Cand.x - p2Cand.x, 2) + Math.pow(p1Cand.y - p2Cand.y, 2),
+        )
+        if (dist < minDistance) {
+          minDistance = dist
+          bestP1 = p1Cand
+          bestP2 = p2Cand
+        }
+      }
+    }
+    return { pointsToConnect: [bestP1, bestP2] }
```
seveibar: so this general idea is fine, but what you're doing is `_findBestConnectionPointsFromDisjointSets`

PR #339, opened 2025-11-25
Add assignable off-board teleporter routing support

0hmX:
**General Idea**

- Implemented a robust system for routing to assignable off-board obstacles. This was achieved by creating dedicated synthetic "anchor" nodes at the center of each assignable obstacle. These nodes are given a special `offboard-port-` name prefix, allowing the pathfinding algorithm to identify them and apply a heavy cost discount, treating them as routing shortcuts. This new "always create" strategy replaces a less reliable implementation that attempted to find and reuse existing nodes.

**Changes to Existing Solvers**

- `AutoroutingPipelineSolver`: Refactored the pipeline logic to always create synthetic nodes for off-board connections, which guarantees they are named correctly for the pathfinder's cost-reduction heuristic to apply. Removed the complex logic that incorrectly deleted connections in non-portal scenarios.
- `CapacityMeshNodeSolver1` & `CapacityMeshNodeSolver2_NodesUnderObstacles`: Simplified these solvers by removing the now-obsolete logic for tracking obstacle IDs on nodes.
- `CapacityPathingSolver` & `CapacityPathingSolver5`: Simplified the logic for identifying portal nodes to rely solely on the `offboard-port-` name prefix, removing the dependency on the old tracking properties.
- `StrawSolver`: Updated to no longer propagate the obsolete obstacle tracking properties to new "straw" nodes.
- `srj-types.ts`: Centralized new branded types (`ObstacleId`, `LayerName`, `GroupId`) for improved type safety and code clarity across the project.
- `isAllInSet.ts`: Added a new generic utility function to check if all items in an array exist in a Set, improving code reuse.

**New Solver**

- `OffboardConnectionEdgeBuilderSolver`: This new solver is now the central component for handling off-board connections. It scans for assignable obstacles, creates a new synthetic anchor node for each, and then builds "bridge" edges to the main grid and "portal" edges between anchors of the same group.

### DIFF
--- a/lib/solvers/AutoroutingPipelineSolver.ts
+++ b/lib/solvers/AutoroutingPipelineSolver.ts
@@ -154,10 +154,10 @@ export class AutoroutingPipelineSolver extends BaseSolver {
     if (!srj || paths.length === 0) return
     const connections = srj.connections
     const existing = new Set(connections.map((conn) => conn.name))
-    const templateByBase = new Map<string, SimpleRouteConnection>()
+    const baseConnectionTemplates = new Map<string, SimpleRouteConnection>()
     for (const conn of connections) {
-      if (!conn.name.includes("#")) {
-        templateByBase.set(conn.name, conn)
+      if (!conn.isPortalConnection) {
+        baseConnectionTemplates.set(conn.name, conn)
       }
     }
     const nodeMap = new Map(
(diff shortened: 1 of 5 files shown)


lib/solvers/AutoroutingPipelineSolver.ts
```diff
@@ -122,8 +126,80 @@ export class AutoroutingPipelineSolver extends BaseSolver {
...
+   */
+  private stripOffboardEdges(stage: string) {
+    if (!this.capacityEdges) return
+    const originalCount = this.capacityEdges.length
+    this.capacityEdges = this.capacityEdges.filter(
+      (edge) => !edge.capacityMeshEdgeId.startsWith("offboard-"),
+    )
+    const removed = originalCount - this.capacityEdges.length
+    this.capacityEdgesWithPortals = null
+  }
```
seveibar: don't pollute top level functions/classes. This function is too specific to be in this class- this pipeline is vary important so we should try to keep utility functions out of this file

lib/solvers/CapacityPathingSolver/CapacityPathingSolver.ts
```diff
@@ -359,6 +357,22 @@ export class CapacityPathingSolver extends BaseSolver {
     this.visitedNodes!.add(currentCandidate.node.capacityMeshNodeId)
   }
 
+  private isNodeTraversable(
+    node: CapacityMeshNode,
+    connectionName: string,
+  ): boolean {
+    if (node.capacityMeshNodeId.startsWith("offboard-port-")) {
```
seveibar: never attach semantics to ids, just have a boolean on the capacity node

seveibar (changes requested):
the AssignableViaAutoroutingPipeline should be modified not the main autorouting pipeline- i realize the name is bad- but yea the AssignableViaAutoroutingPipeline has less stages and specializes for assignable obstacles. I'm not sure if the main pipeline needs to have assignable support (yet?)

PR #355, opened 2025-12-01
Fix #354 Go to Iteration no longer restricted to not solved and not failed

### LINKED ISSUE #354: Go to Iteration Doesn't work Once Solved
Go to Iteration mostly consistently goes to the iteration you tell it to. Specifically, it consistently restarts routing when you tell it to go back in time. When solved, pressing Go to Iteration doesn't even show the prompt() UI.

Heinrich-XIAO:


seveibar:
@Heinrich-XIAO i'm not sure i understand, how can you step on a solver that is failed or solved?

PR #393, opened 2025-12-09
fix: prevent false obstacle rejection by linking MST sub-routes to parent   nets

0hmX:
The autorouter was incorrectly rejecting valid paths as "obstacle
  intersections." This happened because when a net was split into Minimum
  Spanning Tree (MST) segments (e.g., net_mst0), the connectivity solver didn't
  link these sub-segments back to the original net name (e.g., net) used by
  destination pads.

  Solution:
   1. Implemented parentNetId property propagation through the entire solver
      pipeline (NetToPointPairsSolver, CapacityPath, SimpleRouteConnection,
      etc.).
   2. Updated ConnectivityMap generation to explicitly link MST sub-routes to
      their parentNetId and mergedConnectionNames.
   3. Removed fragile regex-based name parsing (_mst stripping) in favor of
      this property-based approach.

  This ensures the solver correctly identifies destination pads as valid
  connections, not obstacles.


seveibar (commented):
everything about this is good except for the name. `originalConnectionName` or `rootConnectionName` is far better, net id doesn't exist as a term and increases the complexity of understanding the autorouter

PR #438, opened 2025-12-22
refactor: consolidate redundant layer transition features in getIntraNodeCrossings

0hmX:
Nutshell: **We were counting the layer transition twice**;

- Simplified getIntraNodeCrossings: Removed the redundant numTransitions return value and the manual numEntryExitLayerChanges counter. Both are now consolidated into a single numEntryExitLayerChanges field derived directly from transitionPairPoints.length.

- Updated MultiHeadPolyLineIntraNodeSolver: Aligned the solver with these changes by renaming its internal usage of numTransitions to numEntryExitLayerChanges.

- Improved Maintainability: The logic now has a single source of truth for layer-change counts, reducing the risk of divergent values.


seveibar (approved):
this looks fine but i'm trying to understand your description, isn't this a noop change? Like the functionality is the same? I don't understand in the description how you said we're counting something twice- it seems like this is just a refactor to remove numTransitions?

0hmX:
@seveibar I also removed a variable called `numEntryExitLayerChanges`, which is the same as `transitionPairPoints.length`, and also removed the extra `numTransitions` from the return type

PR #444, opened 2025-12-24
feat: Introduce SingleTransitionIntraNodeSolver

ShiboSoftwareDev:
This pull request introduces a new, dedicated solver, SingleTransitionIntraNodeSolver, to correctly handle the    
specific case of a single intra-node route that requires a layer transition.                                      

Problem                                                                                                           

The existing solver pipeline lacked a specialized, efficient solver for the simple case of one route needing a    
single via. This could lead to failures or suboptimal solutions, as demonstrated by the pipeline1-bug1            
reproduction case.                                                                                                

Solution                                                                                                          

 • New SingleTransitionIntraNodeSolver: A new, simple solver is created to handle a single route with a layer     
   transition. It correctly places a via with an appropriate margin from the node boundaries.                     
 • HyperSingleIntraNodeSolver Integration: The new solver is integrated into the hyper-parameter supervisor under 
   the closedFormSingleTrace strategy, following the established pipeline conventions.                            
 • Bug Fix & Testing:                                                                                             
    • Adds a new fixture (pipeline1-bug1.fixture.tsx) to debug the failing case.                                  
    • Enables the previously skipped test for this bug and adds a visual snapshot check to prevent regressions.   

This change introduces a more robust and maintainable approach for handling simple layer transitions within the   
high-density routing pipeline. 

before:
Screenshot 2025-12-24 at 9 15 39 AM
after:
Screenshot 2025-12-24 at 9 15 07 AM

### DIFF
--- a/lib/solvers/HighDensitySolver/SingleTransitionIntraNodeSolver.ts
+++ b/lib/solvers/HighDensitySolver/SingleTransitionIntraNodeSolver.ts
@@ -45,6 +45,11 @@ export class SingleTransitionIntraNodeSolver extends BaseSolver {
     }
 
     const route = this.routes[0]
+    if (route.A.z === undefined || route.B.z === undefined) {
+      this.failed = true
+      this.error = `Route points should have predefined z values`
+      return
+    }
     if (route.A.z === route.B.z) {
       this.failed = true
       this.error = "Only one route provided, but it has no transition"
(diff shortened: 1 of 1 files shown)


seveibar (approved):
Nice, cc @0hmX to take a look

lib/solvers/HighDensitySolver/SingleTransitionIntraNodeSolver.ts
```diff
@@ -0,0 +1,218 @@
...
+    this.routes = this.extractRoutesFromNode()
+    this.bounds = this.calculateBounds()
+
+    if (this.routes.length !== 1) {
+      this.failed = true
+      this.error = `Expected 1 route, but got ${this.routes.length}`
+      return
+    }
+
+    const route = this.routes[0]
```
0hmX: I think you need to add another check because z is optional
```
if(route.A.z === undefined || route.B.z === undefined) {
      this.failed = true
      this.error = `Route points should have predefined z values`
      return
    }
```

0hmX (changes requested):

PR #453, opened 2025-12-26
Improve Keepout Solver

seveibar:
image

- **improve keepout, keepout now varies trace order to get better results on subsequent runs with varied order**
- **accept srj outline/bounds**
- **pass srj**
- **add outline push support**
- **filter to route segments near point**
- **increase sample interval**
- **computeDraw revision (1) BREAKS THINGS**
- **wip more improvements to draw position calculation**
- **high density repro**
- **repro problem where line isn't away from edges**
- **fix bounds bug with simple high density route solver**
- **snapshot update**


lib/solvers/TraceKeepoutSolver/TraceKeepoutSolver.ts
```diff
@@ -492,6 +512,70 @@ export class TraceKeepoutSolver extends BaseSolver {
...
+
+    // Create a route for each edge of the outline (on all layers)
+    // We create separate routes for each edge so the spatial index can find them efficiently
+    // Each route needs a unique connection name for the spatial index
+    for (let i = 0; i < outlinePoints.length; i++) {
+      const start = outlinePoints[i]!
+      const end = outlinePoints[(i + 1) % outlinePoints.length]!
+
+      // Create route on z=0 (top layer)
+      routes.push({
```
0hmX: Are we adding this only in the top and bottom layer?

PR #454, opened 2025-12-26
update rectdiff to use the new gapfill solver

0hmX:


seveibar (commented):
ooh we gotta get this guy merged... but the snapshots have all changed 😖

0hmX:
This is an old one sorry for the confusion. @seveibar this is the real one we already merged https://github.com/tscircuit/tscircuit-autorouter/pull/455

PR #455, opened 2025-12-26
update rectdiff to use gapfill

0hmX:


seveibar (approved):
probably took a bit for the snapshots to update 😁

PR #465, opened 2025-12-29
rename examples to fixture

0hmX:
- **WIP: working**
- **update to using with from assert**
- **format**


seveibar (approved):
test failures

PR #500, opened 2026-01-05
handle more cases and get more sanitized data for the input params

0hmX:
This is handling the case where JSON.Stringly will fail like Maps and Sets, and also check for recursive objects


lib/testing/AutoroutingPipelineDebugger.tsx
```diff
@@ -45,6 +45,59 @@ const PIPELINE_SOLVERS = {
 const PIPELINE_STORAGE_KEY = "selectedPipeline"
 const EFFORT_STORAGE_KEY = "selectedEffort"
 
+const sanitizeParamsForDownload = (
```
imrishabh18: I see a lot of `any` being used in this method, are the input params arbitrary?

0hmX: This is handling the case where JSON.Stringly will fail like Maps and Sets, and also check for recursive objects

PR #503, opened 2026-01-06
broken SectionPath, with multiple entry and exit are treated as single SectionPath

0hmX:
When a path enters and exits the section multiple times, all segments are into a single SectionPath that spans from the first entry to the last exit. This ensures the entire affected portion is re-routed as one unit.


seveibar (approved):
@0hmX NICE WORK

PR #506, opened 2026-01-07
feat: Add straight line deviation penalty to improve routing quality

ShiboSoftwareDev:
This change adds a new hyperparameter `STRAIGHT_LINE_DEVIATION_FACTOR` to the PortPointPathingSolver that penalizes paths that deviate significantly from the straight line between connection endpoints. This encourages more direct routing and reduces unnecessarily complex paths.

Key improvements:
- Routes that deviate from the optimal straight line path are penalized in the A* heuristic
- Reduces excessive path complexity (e.g., 33+ points reduced to 5 points for vertical connections, in bugreport28)

The best penalization factor I found was 4, providing good routing quality improvements while avoiding over-constraining the algorithm.

There were also some performance improvements in some tests:
before:
Screenshot 2026-01-06 at 3 16 26 PM
after:
Screenshot 2026-01-06 at 3 16 44 PM

bugreport8 is now solver, before:
Screenshot 2026-01-07 at 7 48 55 AM
after:
Screenshot 2026-01-07 at 7 49 05 AM

### DIFF
--- a/lib/solvers/PortPointPathingSolver/PortPointPathingSolver.ts
+++ b/lib/solvers/PortPointPathingSolver/PortPointPathingSolver.ts
@@ -58,7 +58,7 @@ export interface PortPointPathingHyperParameters {
   JUMPER_PF_FN_ENABLED?: boolean
 
   /** Factor for penalizing deviation from straight line path */
-  STRAIGHT_LINE_DEVIATION_FACTOR?: number
+  STRAIGHT_LINE_DEVIATION_PENALTY_FACTOR?: number
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/solvers/PortPointPathingSolver/PortPointPathingSolver.ts
```diff
@@ -788,11 +795,62 @@ export class PortPointPathingSolver extends BaseSolver {
...
+      yy = lineEnd.y
+    } else {
+      xx = lineStart.x + param * C
+      yy = lineStart.y + param * D
+    }
+
+    const dx = point.x - xx
+    const dy = point.y - yy
+    return Math.sqrt(dx * dx + dy * dy)
+  }
```
seveibar: avoid instance methods like this at all costs, wastes very valuable algorithm LoC

this is a builtin to math-utils, you can import from there

seveibar (commented):
this is quite strange, i'm checking some data on whether or not this improves outcomes with our baseline (which is in the first test node)

image

seveibar (approved):

PR #516, opened 2026-01-08
Make SameNetViaMerger  run after path simplification steps

### LINKED ISSUE #369: SameNetViaMerger needs to run after path simplification steps

Excellencedev:
/fixes #369


seveibar (commented):
the fix is ok but we don't have any snapshots showing a difference, we kind of need a test

Excellencedev:
unstale

PR #517, opened 2026-01-08
update params for MultiSectionSolver

0hmX:
because we are now stable with lower values of expansion degrees i was think if we could use lower values of expansion degrees to get the output also reduce MAX_SECTION_ATTEMPTS = 50


seveibar:
@0hmX higher EXPANSION_DEGREEs will always perform better if given enough time. So we use the user's effort level to determine the amount of time that should be spent.

Expansion degrees over 10 are generally unlikely to improve too much. We can also use gradient descent to learn the correct values if we construct a dataset

PR #543, opened 2026-01-16
better redistribution + faster using better data structure

0hmX:
- **easier lookups**
- **add better z check logic**
- **sort to preserver order of port pots**
- **better visualization**
- **update snapshot test**


seveibar (approved):
Why dint the snapshot change on the tests i mentioned in my video?

PR #581, opened 2026-02-24
implement bfs check

0hmX:
WIP

### DIFF
--- a/lib/solvers/AvailableSegmentPointSolver/AvailableSegmentPointSolver.ts
+++ b/lib/solvers/AvailableSegmentPointSolver/AvailableSegmentPointSolver.ts
@@ -21,7 +21,7 @@ export interface SegmentPortPoint {
   /** XY distance to the centermost port on this Z level (centermost port has distance 0) */
   distToCentermostPortOnZ: number
   /**
-   * This is special port points that exists to be used when reaching an obstacle is super impossible
+   * This is special port point that is created in narrow gaps, and only kept when otherwise reaching the obstacle is impossible
    * ideally this port points should be discarded but we need them in some cases
    */
   cramped: boolean


lib/solvers/AvailableSegmentPointSolver/AvailableSegmentPointSolver.ts
```diff
@@ -20,6 +20,11 @@ export interface SegmentPortPoint {
   rootConnectionName?: string
   /** XY distance to the centermost port on this Z level (centermost port has distance 0) */
   distToCentermostPortOnZ: number
+  /**
+   * This is special port points that exists to be used when reaching an obstacle is super impossible
+   * ideally this port points should be discarded but we need them in some cases
```
seveibar:
```suggestion
   * This is special port point that is created in narrow gaps, and only kept when otherwise reaching the obstacle is impossible
   * ideally this port points should be discarded but we need them in some cases
```

lib/solvers/HopCheckSolver/solver/FindCrampedPortPointsToMakeUnreachableRegionsContainingObstacleReachableSolver.ts
```diff
@@ -0,0 +1,111 @@
...
+
+type FindCrampedPortPointsToMakeUnreachableRegionsContainingObstacleReachableSolverInput =
+  HopCheckSolverInput & {
+    regionsWithObstacle: TypedRegion[]
+  }
+
+/**
+ * Finds cramped port points that can potentially make unreachable obstacle regions reachable.
+ */
+export class FindCrampedPortPointsToMakeUnreachableRegionsContainingObstacleReachableSolver extends BaseSolver {
```
seveibar: there are better names than this

lib/solvers/HopCheckSolver/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  d: CapacityMeshNode
+}
+export type TypedRegionPort = Omit<RegionPort, "d"> & {
+  d: SegmentPortPoint
+}
+
+export type TypedHyperGraph = Omit<HyperGraph, "ports" | "regions"> & {
+  ports: TypedRegionPort[]
+  regions: TypedRegion[]
+}
```
seveibar: you can add this to the HyperGraph repo and the HyperGraph definition, no need to define this here

PR #586, opened 2026-02-25
Fix benchmarking workflow for forked repo prs

MustafaMulla29:


.github/workflows/benchmark.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         github.event.comment.author_association == 'COLLABORATOR'
       )
     runs-on: ubuntu-latest
-    timeout-minutes: 360
+    timeout-minutes: 30
```
0hmX: Is the timeout change necessary? I set it to 360, but sometimes it takes two hours or more to run. 30 is too low.

MustafaMulla29: Oh sorry, did not notice I changed the timeout, @0hmX feel free to update

PR #587, opened 2026-02-25
Fix incorrect intra-node via selection by using region-scoped nearest vias and actual bottom-layer paths (eliminates misattributed via transitions)

AnasSarkiz:


lib/solvers/FixedTopologyHighDensityIntraNodeSolver/FixedTopologyHighDensityIntraNodeSolver.ts
```diff
@@ -2,14 +2,17 @@ import {
   type XYConnection as HgXYConnection,
   type JPort,
   type JRegion,
+  type ViaData,
```
seveibar: remember, data is a banned word. Should never be in the codebase

PR #589, opened 2026-02-26
update PF

0hmX:


seveibar:
@0hmX so even though it isn't obvious, did it work?

seveibar:
@0hmX @MustafaMulla29 does the benchmark run in parallel? I'm happy to get some souped up machines for this if needed

PR #592, opened 2026-02-27
Swap via diameter to 0.3mm

seveibar:


MustafaMulla29:
> @MustafaMulla29 the benchmark command should run _all solvers_ by default. It's currently running a "default solver" which I don't think makes sense

So if you don't specify a name then it should run all right?

seveibar:
> > @MustafaMulla29 the benchmark command should run _all solvers_ by default. It's currently running a "default solver" which I don't think makes sense
> 
> So if you don't specify a name then it should run all right?

@MustafaMulla29 yep!

PR #610, opened 2026-03-02
Introduce FixedTopologyHighDensityIntraNodeSolver to HyperSingleIntraNodeSolver to handle dense intra-node routing with a fixed-topology path

AnasSarkiz:


seveibar:
@MustafaMulla29 i know you already did this and i broke it- but we need to show the results from main in a details tab so that we know if there's improvement

seveibar:
+1.1% (main shown below)
image

PR #863, opened 2026-04-10
Update high-density-repair02: repaired traces stay inside boundaries by nudging only interior points inward

AnasSarkiz:


seveibar:
can you include in your title what improved, what feature or bug was fixed etc.?

seveibar (approved):

AnasSarkiz:
> can you include in your title what improved, what feature or bug was fixed etc.?

added

PR #872, opened 2026-04-11
Integrate HighDensityForceImproveSolver

ShiboSoftwareDev:


seveibar (approved):
This is from the arduino uno- i think we do need to get this time wayy down. If we want to merge this i think it's ok, but we will need to very quickly improve the performance, because it is really slow!

image

PR #891, opened 2026-04-13
Add pour-via-escape06 fixture and regression test

seveibar:
## Summary
- add a new `pour-via-escape06` feature fixture from bug report `1b2d3e71-2a60-4e58-8415-c8bfa85e4561`
- add a debugger fixture wired to the new `simple_route_json`
- add a snapshot-backed regression test covering the 4-layer pour-via-escape case with `bottom` and `inner2` copper pours

## Testing
- `bun test tests/features/pour-via-escape/pour-via-escape06.test.ts`


seveibar:
@0hmX this is a repro i'd like you to take a look at! The nodes are availableZ=[0] instead of availableZ=[0,1] and that causes issues

PR #902, opened 2026-04-14
chore update core package version

mohan-bee:
Older @tscircuit/core drops the custom trace thickness/width before building autorouter input, so the autorouter never receives nominalTraceWidth and falls back to the default trace width.


mohan-bee:
need help ! why these tests are failing ?

MustafaMulla29:
@tscircuit/math-utils is outdated somewhere

PR #916, opened 2026-04-16
switch rectdiff from npm to git url

0hmX:
- **update to git link for rectdiff**
- **update to next commit**

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@tscircuit/hypergraph": "^0.0.71",
     "@tscircuit/jumper-topology-generator": "^0.0.4",
     "@tscircuit/math-utils": "^0.0.27",
-    "@tscircuit/rectdiff": "0.0.26",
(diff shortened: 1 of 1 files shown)


0hmX:
@seveibar

0hmX:
This might not be ideal, but I think we need to drop a few commits in rectdiff because they’re introducing bugs. The autorouter is still on **version 0.0.26**, while rectdiff has moved ahead to 0.0.29.

I tried using the latest version, but it’s becoming unstable in both the CLI and Node. A lot has changed there—especially how port points are created—which is likely causing the issues.

At this point, I’m not confident about the recent commits. I do want to debug the problem, but it’s turning into a real pain.

So for now, I’m planning to fork it and **roll back to a commit equivalent to 0.0.26**, then debug from there. That’s what this commit is for. We don’t need to merge it—this is just to unblock progress and figure things out.

seveibar:
@0hmX sounds good!

PR #938, opened 2026-04-20
corner port points

0hmX:
A fixture to visualize cornet port points, highlight the used corner port points


lib/autorouter-pipelines/AutoroutingPipeline4_TinyHypergraph/AutoroutingPipelineSolver4_TinyHypergraph.ts
```diff
@@ -478,7 +478,13 @@ export class AutoroutingPipelineSolver4_TinyHypergraph extends BaseSolver {
 
   solveUntilPhase(phase: string) {
     while (this.getCurrentPhase() !== phase) {
+      if (this.failed) {
+        return
+      }
       this.step()
+      if (this.failed) {
+        return
+      }
```
seveibar, line 487: This is wasting some cpu- you can move this second check outside the loop, not a huge deal

seveibar (commented):
Hmm seems to cause lots of failures- do we have an alternate solution? Should corners be removed? I think the problem isnt corner points, but single corner points when regions are bordering on corners- but maybe im wrong…. Im not sure but worried about the new test failures- maybe we can maybe it less aggressive as a check

PR #950, opened 2026-04-21
Honor explicit via pad/hole dimensions across autorouters

ShiboSoftwareDev:
Normalize SRJ via dimensions through a shared helper so all autorouting pipelines and solvers use min_via_pad_diameter ahead of deprecated minViaDiameter. Pipeline4 now also propagates the requested hole diameter into simplified routes.

  Update circuit-json export and SVG rendering to preserve and draw the routed via size correctly, including via_hole_diameter on emitted vias.


0hmX:
@ShiboSoftwareDev  You need to rebase against https://github.com/tscircuit/tscircuit-autorouter and then force push.Just updating snapshots will not work

seveibar (approved):

PR #953, opened 2026-04-21
add focused repro for high-density solver issue caused by reentry in nodeWithPortPoints input

0hmX:
image

current
image


seveibar (approved):
Nice!!!!!!!!!!

PR #959, opened 2026-04-21
update url for rectdiff

0hmX:


seveibar (approved):
Can you push to tscircuit branches so that youre not linking to a fork on your org. Make sure to inform the discord channel if a change causes drc to drop,

PR #976, opened 2026-04-24
Skip zero-length trace segments from duplicate route points

mohan-bee:
This PR prevents zero-length trace segments from being emitted when autorouter route data contains consecutive duplicate points.

Previously, convertHdRouteToSimplifiedRoute preserved adjacent same-layer duplicates such as A -> A -> B, which could later become a degenerate A -> A copper segment in exported PCB output. That kind of ghost segment can trigger downstream DRC issues even when the visible routing looks correct.


seveibar:
@mohan-bee what motivates this fix?

seveibar:
it seems fine but i'm curious how you found it, why it is a problem etc.

mohan-bee:
> it seems fine but i'm curious how you found it, why it is a problem etc.

@seveibar  https://github.com/tscircuit/circuit-json-to-kicad/pull/242 

first i thought that is from kicad and then @imrishabh18  caught the core issue !

seveibar (approved):

PR #1047, opened 2026-04-28
add approximating rects to srj  AutoroutingPipeline4

Sang-it:


lib/autorouter-pipelines/AutoroutingPipeline4_TinyHypergraph/AutoroutingPipelineSolver4_TinyHypergraph.ts
```diff
@@ -52,6 +52,7 @@ import { SingleLayerNodeMergerSolver } from "../../solvers/SingleLayerNodeMerger
 import { StrawSolver } from "../../solvers/StrawSolver/StrawSolver"
 import { TraceSimplificationSolver } from "../../solvers/TraceSimplificationSolver/TraceSimplificationSolver"
 import { TraceWidthSolver } from "../../solvers/TraceWidthSolver/TraceWidthSolver"
+import { covertSrjToOldFormat } from "lib/utils/covertSrjToOldFormat"
```
seveibar: never use "old" or "legacy". Be specific, otherwise we'll have "old2" "2026" and a bunch of other bad descriptions

in this case `addApproximatingRectsToSrj` is appropriate

seveibar (changes requested):
need to improve naming, then it looks ok

seveibar (approved):

PR #1071, opened 2026-04-30
Update the checks to latest with new drc checks added

imrishabh18:


tests/drc-via-spacing.test.ts
imrishabh18, line 1: This test were failing and had to be modified, cause the previous measurement of checks were from the via pad edges and not the hole edges

PR #1093, opened 2026-05-04
feat: add overrides option for custom rectdiff ONLY AutoroutingPipelineSolver4_TinyHypergraph

0hmX:
Expose rectdiff through a new overrides option so callers can pass a custom rectdiff implementation. This makes it easier to test inside the rectdiff solver repo by using a local rectdiff version.


seveibar:
Idk if i like this bc it breaks the common pattern BUT you can still use the changes by clicking on the build workflow and using the custom tar gz temporarily?

seveibar:
Its not a bad idea but i want to think about how this could be a bit more ergonomic

PR #1118, opened 2026-05-05
Add geometry-aware DRC checks for rotated pads and reconstructed obstacle geometry

Abse2001:


lib/testing/getDrcErrors.ts
```diff
@@ -46,6 +52,21 @@ export const getDrcErrors = (
   const traceErrors = checkEachPcbTraceNonOverlapping(circuitJson, {
     minClearance: options.traceClearance,
   })
+  const geometryAwarePadIds = getGeometryAwarePadIds(circuitJson)
+  const geometryAwareTraceErrors = getGeometryAwarePadTraceErrors(
```
seveibar: aren't all pad trace errors geometry aware? This doesn't seem like the right idea- shouldn't we be fixing the existing errors?

seveibar (commented):
is this a hack to avoid fixing tscircuit/checks?  all pad errors should be aware of rotation- is it because of flatbush or something?

lib/types/srj-types.ts
```diff
@@ -74,6 +74,12 @@ export interface Obstacle {
   height: number
   /** Optional counter-clockwise rotation metadata in degrees. */
   ccwRotationDegrees?: number
+  originalRotatedRect?: {
+    center: { x: number; y: number }
+    width: number
+    height: number
+    ccwRotationDegrees: number
+  }
```
seveibar: this format is really nice, doesn't have many hacks and is used by partners. Let's not add hacks to it

you can just use the obstacles inside originalSrj or something? The originalSrj contains rotations

seveibar (changes requested):
i think this is a hack for it?

PR #1121, opened 2026-05-06
Prevent downstream crashes by adding x/y to through_obstacle route points

AnasSarkiz:


seveibar (changes requested):
ah sorry, didn't know this is what you meant- this is confusing and we should fix downstream

PR #1123, opened 2026-05-06
Preserve simplified path prefixes when 45-degree completion fails

Abse2001:


seveibar:
Hmm we should have a snapshot test ig, seems to do no harm?

imrishabh18 (approved):

PR #1129, opened 2026-05-06
0hmx/use rectdiff patch out of bounds

0hmX:


seveibar:
@0hmX you aren't allowed to change the dataset

0hmX:
continue in https://github.com/tscircuit/tscircuit-autorouter/pull/1134

Issue #66, opened 2025-03-20
Implement trace thickness as a parameter


seveibar:
You must have a major contribution to this repo prior to attempting this.

I think this issue is really hard and the bounty is likely to increase over time. There are a lot of implications of trace thickness for high density routing and probability calculation etc.

I'm not even really sure anyone should try to attempt to get trace thickness to go in a single PR, it probably will take a lot of effort.

We mainly only care about trace thickness multiples (which might help with the implementation). I.e. right now every trace is the industry standard data line thickness `0.15mm`, but we really only need to add support for 2x, 4x and 8x multiples, like `0.3mm`, `0.6mm`  and `1.2mm`

In many cases, thickness isn't really adequate to specify what the user wants. The user wants to make sure enough power can flow to particular chips. This can be done by creating more traces to the chip, or by routing thicker lines. Routing more traces is "autorouter friendly", we're more likely to succeed because we can weave the trace through obstacles. Thick traces are more "human friendly", it's easy to see what carries power.

In practice I believe having good trace adjacency behavior (combining traces in the same net when they're nearby) is a bit better than thickness, and maybe a good idea to do prior to this

/bounty $150

CuboYe:
Execution refresh from Tez (Mar 14):

- Issue remains OPEN with active bounty and prior territory claim from contributor side: https://github.com/tscircuit/tscircuit-autorouter/issues/66#issuecomment-4025066093
- Existing implementation PR is still open and directly claims this issue: https://github.com/tscircuit/tscircuit-autorouter/pull/638
- PR #638 currently shows green checks (Vercel + benchmark instructions) and benchmark follow-up comments by author.

Given the active overlapping implementation, I am not opening a duplicate competing PR in this cycle to avoid reviewer bandwidth waste.

If maintainers want a second independent implementation/review lane, I can start immediately and post a scoped first-attempt PR in the next cycle.

CuboYe:
Post-rp2040 pivot activation: I’m taking a focused next-attempt lane on #66 with a test-first slice.

Execution plan (unit-test aligned):
1) Add a narrow failing regression test covering trace_thickness propagation through the routing path (2x/4x/8x representative cases).
2) Implement minimal fix to ensure thickness parameter reaches route generation consistently.
3) Run targeted tests + full suite subset, then open PR with `/claim #66`.

[...]

PR #1155, opened 2026-05-12
Support top-level nominal trace width

shengtenghou4-star:
/claim #66

### DIFF
--- a/lib/autorouter-pipelines/AssignableAutoroutingPipeline2/AssignableAutoroutingPipeline2.ts
+++ b/lib/autorouter-pipelines/AssignableAutoroutingPipeline2/AssignableAutoroutingPipeline2.ts
@@ -393,6 +393,7 @@ export class AssignableAutoroutingPipeline2 extends BaseSolver {
         connMap: cms.connMap,
         colorMap: cms.colorMap,
         minTraceWidth: cms.minTraceWidth,
+        nominalTraceWidth: cms.srj.nominalTraceWidth,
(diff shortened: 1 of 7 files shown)


shengtenghou4-star:
Implementation notes:\n- Adds top-level SimpleRouteJson nominalTraceWidth as the default trace-width target for TraceWidthSolver.\n- Keeps per-connection nominalTraceWidth as the higher-priority override.\n- Passes the top-level default through the active autorouting pipelines that instantiate TraceWidthSolver.\n- Adds regression coverage for top-level default behavior and per-connection override behavior.\n\nVerification:\n- npx bun test tests/features/tracewidthsolver/tracewidthsolver-default-nominal-width.test.ts tests/features/tracewidthsolver/tracewidthsolver01.test.ts tests/features/tracewidthsolver/tracewidthsolver02.test.ts\n- npx bun x tsc --noEmit\n- git diff --check\n\nNote: full biome check on this Windows checkout wants to rewrite existing CRLF line endings/import order in broad existing files, so I only ran targeted lint without formatter/organize-imports on the changed solver/test to avoid noisy unrelated churn.

shengtenghou4-star:
CI follow-up:
- Fixed compatibility with legacy SRJ files that set nominalTraceWidth: null; TraceWidthSolver now only treats finite numeric widths as active overrides/defaults.
- Added a regression test for null handling.
- GitHub checks are now green for build, type-check, format-check, Vercel, and Bun test shards 1-5.

PR #1167, opened 2026-05-14
Fix same-root MST route disconnections in repro117

ShiboSoftwareDev:
Repairs disconnected intra-node route fragments when a same-root MST sibling already
  contains the missing local bridge, preserving connectivity for cases like cmn_68 in
  repro117.


lib/solvers/HyperHighDensitySolver/HyperSingleIntraNodeSolver.ts
```diff
@@ -21,6 +22,136 @@ import {
...
+        repairedRoutes,
+        connectionName,
+        pointKey(portPoints[0]!),
+      )
+    }
+  }
+
+  return repairedRoutes
+}
+
```
seveibar, line 24: can you move these into a different file? This file is important, we should avoid pollution

seveibar (approved):
This is fine but please move the repair function to a separate file!!!

PR #1178, opened 2026-05-15
fix: stitch solver disconnecting the bridge fragment of the same root net

imrishabh18:
## Before
image

## After
image


seveibar:
CC @0hmX

seveibar (approved):

0hmX (approved):

PR #1185, opened 2026-05-16
Fix same-root shared junction routing

ShiboSoftwareDev:
Handles exact same-root shared junctions by collapsing duplicate physical junctions only
  for high-density solving, while preserving original connection memberships for final route
  repair.

  Guards restored same-root fallback routes with a local clearance check so they are not
  emitted when they would collide with different-root routed segments.


seveibar (commented):
bugreport44 needs to be fixed, that pad isn't connected now

PR #1201, opened 2026-05-21
devdeps: add missing connectivity-map, polished, and transformation-matrix

itisrohit:
### Summary

This PR adds three direct dependencies to `devDependencies` that are actively used in the codebase but were missing from `package.json`:

1. **`connectivity-map`**
   * Imported in:
     * `lib/solvers/RouteStitchingSolver/MultipleHighDensityRouteStitchSolver.ts`
     * `lib/solvers/RouteStitchingSolver/MultipleHighDensityRouteStitchSolver3.ts`
     * `lib/autorouter-pipelines/AssignableAutoroutingPipeline2/RelateNodesToOffBoardConnectionsSolver.ts`

2. **`polished`**
   * Imported in:
     * `lib/solvers/colors.ts`
     * `lib/utils/getPresuppliedTraceVisualization.ts`

3. **`transformation-matrix`**
   * Imported in:
     * `lib/solvers/JumperPrepatternSolver/JumperPrepatternSolver2_HyperGraph.ts`
     * `lib/solvers/UnravelSolver/CachedUnravelSectionSolver.ts`

### Why this is needed

Relying on transitive dependency hoisting can lead to unpredictable build failures when building in strict environments, fresh clones, or under package managers like `pnpm`. Listing these directly inside `devDependencies` ensures the build works correctly on all setups without relying on parent/transitive hoisting.

### Verification

* Verified that the package successfully compiles via `bun run build`.
* Ran the SVG snapshot test suite locally (`bun test tests/svg.test.ts`) and verified it passes cleanly.


imrishabh18 (changes requested):
I don't think we need this, are you facing any build issues? I don't see build failing on `main`

itisrohit:
Hi @imrishabh18!

Ah, got it! The build on main passes fine because they're already pulled in by our other dependencies. 

However, since we import them directly in a few files (like `connectivity-map` in the route stitching solver, `polished` in colors, and `transformation-matrix` in unravel), I thought it might be safer to list them explicitly in `devDependencies` just to prevent any future build breakages if those other dependencies ever update.

But if you feel it's unnecessary noise in package.json, I'm totally fine with closing this PR! Let me know what you prefer.

PR #1203, opened 2026-05-21
Add Pipeline8 for routing the board with pre placed via's

imrishabh18:


seveibar (changes requested):
Glad youre working on this but wrong approach

- Pipelines shouldnt extend other pipelines as a general rule, its too confusing
- this does not feed proper hypergraphs into tiny-hypergraph, this is a post processing fix whivh is not the right stage

seveibar:
To make this work

- copy pipeline4
- before the portpointpathing add a stage that breaks all nodes into constituent single layer nodes
- Modify the hypergraph to handle assignable vias
- run the rest of the pipeline normally

lib/solvers/RouteStitchingSolver/SingleHighDensityRouteStitchSolver3.ts
imrishabh18, line 1: Prevents route fragments with same point but are on different layer were joined into a via in this stitch solver, which was not pre placed

lib/autorouter-pipelines/AutoroutingPipeline8/SingleLayerNodePortPointSolver.ts
```diff
@@ -0,0 +1,152 @@
...
+  sharedEdgeSegments: SharedEdgeSegment[]
+}
+
+const getSingleLayerNodeId = (nodeId: CapacityMeshNodeId, z: number) =>
+  `${nodeId}__z${z}`
+
+const uniqueSorted = (values: number[]) =>
+  [...new Set(values)].sort((a, b) => a - b)
+
+export class SingleLayerNodePortPointSolver extends BaseSolver {
```
seveibar: I don't think this name is clear- SplitNodesIntoSingleLayerNodeSolver is more clear

lib/autorouter-pipelines/AutoroutingPipeline8/getXyPointKey.ts
```diff
@@ -0,0 +1,2 @@
+export const getXyPointKey = (point: { x: number; y: number }) =>
```
seveibar, line 1: move this file/function- keep it in the relevant pipeline, lib/utils is an anti-pattern

lib/autorouter-pipelines/AutoroutingPipeline8/assignableViaUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+    ),
+  ].sort((a, b) => a - b)
+
+export const getAssignableViaPointKey = (obstacle: Obstacle) =>
+  getXyPointKey(obstacle.center)
+
+export const getAssignableViaPointKeys = (obstacles: Obstacle[]) =>
+  new Set(
+    obstacles.filter(isAssignableViaObstacle).map(getAssignableViaPointKey),
+  )
```
seveibar, line 33: move these to the pipeline direction

seveibar (approved):
the results are fantastic, see very important comments

PR #1205, opened 2026-05-21
benchmark qol

ShiboSoftwareDev:
Adds structured benchmark failure summaries so solver failures are separated from relaxed DRC failures. Records DRC error counts/types/messages per sample, adds top solver failure buckets to text/JSON output, and scales sample timeouts by concurrency to avoid misattributing parallel wall-clock contention as phase failures.

### DIFF
--- a/benchmark.sh
+++ b/benchmark.sh
@@ -77,7 +77,7 @@ Options:
   --scenario-limit N   Run only first N scenarios (same as second positional arg)
   --concurrency N      Number of Bun workers used per solver, or "auto"
   --effort N           Override scenario effort multiplier
-  --sample-timeout D   Override per-sample timeout directly; otherwise timeout is 60s + 60s * effort
+  --sample-timeout D   Override isolated per-sample timeout; benchmark scales it by concurrency for wall-clock execution
   --dataset NAME       Dataset to benchmark: 1/dataset01 (default), zdwiel, 5/srj05, 11/srj11, 12/srj12, 13/srj13, 14/srj14, 15/srj15, or 16/srj16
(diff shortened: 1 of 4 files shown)


benchmark.sh
```diff
@@ -77,7 +77,7 @@ Options:
   --scenario-limit N   Run only first N scenarios (same as second positional arg)
   --concurrency N      Number of Bun workers used per solver, or "auto"
   --effort N           Override scenario effort multiplier
-  --sample-timeout D   Override per-sample timeout directly; otherwise timeout is 60s + 60s * effort
+  --sample-timeout D   Override isolated per-sample timeout; benchmark scales it by concurrency for wall-clock execution
```
seveibar: i don't think we should be scaling by concurrency

seveibar: it doesn't make sense- we have concurrency set to number of cpu cores, samples have independent cores

scripts/benchmark/index.ts
```diff
@@ -519,10 +649,15 @@ const formatProgressDetails = (progress?: WorkerProgress) => {
 const executeTaskOnWorker = (
   slot: WorkerSlot,
   request: WorkerTaskMessage,
+  workerCount: number,
   sampleTimeoutMs?: number,
 ): Promise<WorkerExecutionResult> => {
   return new Promise((resolve) => {
-    const taskTimeoutMs = getTaskTimeoutMs(request.task, sampleTimeoutMs)
+    const taskTimeoutBudget = getTaskTimeoutBudget(
```
seveibar: revert

seveibar (commented):
see comments

seveibar (approved):

PR #1216, opened 2026-05-22
tiny-hypergraph "never fail", Pipeline7 performance improvement to 65% on dataset-13

seveibar:
- Better MAX_ITERATIONS for section hypergraph optimizer
- tiny-hypergraph accepts the best solution rather than running out of iterations
- duplicate port solver + adder, duplicate ports are penalized by 1000mm and only used as a backup


seveibar:
/benchmark --dataset 16 --pipeline 7 --effort 5

PR #1218, opened 2026-05-22
fix: Increasing the effort level was preventing globalDrcForceImprovementSolver to use the preplaced via's

imrishabh18:
Increasing the effort is causing the `globalDrcForceImprovementSolver` to find a better drc score layer change transition at different position and not using the pre placed via.

Another solution for this could be to pass optional constraint which will guard `globalDrcForceImprovementSolver` to use only the passed layer change positions. Is that a better solution? @seveibar


seveibar (approved):
I do feel that it is a bit better to pass a flag like dontMoveVias

PR #1230, opened 2026-05-22
fix: stabilize autorouter route finalization edge cases

itisrohit:
## Description

This PR fixes a cluster of related route-finalization failures in the capacity autorouter. The failures surfaced as clearance violations, duplicate same-net vias, simplification dropping required connection points, and simplification stalls on duplicate route points.

These issues all happen along the same late-stage path from escape-via generation through route stitching, same-net via merging, and trace simplification, so the fixes are grouped here as one solver-stability pass rather than separate unrelated changes.

### What Changed

1. **Escape via trace clearance**
   - Fixed centerline-to-edge clearance math in `EscapeViaLocationSolver`.
   - Prevents escape traces from being accepted when they are too close to rectangular pad obstacles.

2. **Same-net via merge deduplication**
   - Added coordinate-based deduplication in `SameNetViaMergerSolver`.
   - Prevents redundant same-net vias from piling up in solver outputs.

3. **Connection-point preservation during simplification**
   - Passed connection-pin data into the simplification path.
   - Marks required connection points as mandatory stops so simplification does not collapse routes past endpoints.

4. **Duplicate-point handling before simplification**
   - Filters consecutive duplicate same-layer coordinates before constructing simplified routes.
   - Avoids zero-length segments that could stall the degree-45 simplifier.

---

### 1. Escape Via Trace Clearance
*In the **Before** state, the trace centerline calculation allowed the via's trace to route too close to the pad. In the **After** state, the clearance check is correctly computed as `trace radius + obstacle margin`, correctly rejecting the violating candidate via.*

| Before | After |
| :---: | :---: |
| Before clearance fix | After clearance fix |

---

### 2. Path Simplification Collapse
*In the **Before** state, the simplifier collapsed the trace loop entirely and dropped the pin connection. In the **After** state, target pin connection coordinates are correctly preserved as mandatory stops.*

| Before | After |
| :---: | :---: |
| Before simplification collapse fix | After simplification collapse fix |

---

## Verification

- Added or updated targeted regression coverage for the affected solver stages.
- Updated SVG snapshots where route output changed intentionally.
- Verified CI checks are passing for build, type-check, format-check, Vercel build, and all Bun test shards.

Useful local commands:

```bash
bun test tests/bugs/
bun x tsc --noEmit
bun run format:check
```


tests/bugs/bugreport04-aa1d41.test.ts
```diff
@@ -6,10 +6,14 @@ import { getLastStepSvg } from "tests/fixtures/getLastStepSvg"
...
+test(
+  "bugreport04-aa1d41.json-AutoroutingPipeline1_OriginalUnravel",
+  () => {
+    const solver = new AutoroutingPipeline1_OriginalUnravel(srj)
+    solver.solve()
+    expect(getLastStepSvg(solver.visualize())).toMatchSvgSnapshot(
+      import.meta.path,
+    )
+  },
+  { timeout: 30000 },
```
0hmX: don add timeouts its created problem in CI

tests/bugs/bugreport12-35ce1c.test.ts
```diff
@@ -13,7 +13,9 @@ const srj = bugReport.simple_route_json as SimpleRouteJson
 test("provided simple route json snapshot", () => {
   const solver = new AutoroutingPipelineSolver(srj)
   solver.solve()
-  expect(getLastStepSvg(solver.visualize())).toMatchSvgSnapshot(
-    import.meta.path,
-  )
+  const snapshotPath =
+    process.platform === "linux"
```
0hmX: ? is this needed?

tests/bugs/bugreport16-d95f38.test.ts
```diff
@@ -11,7 +11,9 @@ const srj = bugReport.simple_route_json as SimpleRouteJson
 test("bugreport16-d95f38.json", () => {
   const solver = new AutoroutingPipelineSolver(srj)
   solver.solve()
-  expect(getLastStepSvg(solver.visualize())).toMatchSvgSnapshot(
-    import.meta.path,
-  )
+  const snapshotPath =
+    process.platform === "linux"
+      ? import.meta.path.replace(/\.test\.ts$/, "-linux.test.ts")
```
0hmX: ?

tests/bugs/escape-via-trace-clearance.test.ts
```diff
@@ -0,0 +1,54 @@
+import { expect, test } from "bun:test"
+import { EscapeViaLocationSolver } from "lib/solvers/EscapeViaLocationSolver/EscapeViaLocationSolver"
```
0hmX: I don’t think we need this test. It verifies the math, which is useful, but we might change it later, and it’s not important enough to test.

We care about how the final board looks using visual tests.

tests/bugs/simplification-connection-preservation.test.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, test } from "bun:test"
```
0hmX: You can skip non visual tests. no need to test.

tests/__snapshots__/e2e3-multisection-linux.snap.svg
0hmX, line 1: Why are new SVG snapshot files being created? Not sure what new thing are they testing

0hmX (changes requested):

0hmX:
image



your pr description make it sound it like it should be 4 seperate prs

itisrohit:
@0hmX  I updated the title and PR description to frame the changes as a single route-finalization stability pass instead of four separate fixes.

The common scope here is the late-stage autorouter flow:
escape-via generation → route stitching / same-net via merging → trace simplification.

I kept the before/after screenshots, but rewrote the description around that shared pipeline context.

I also addressed the earlier review comments.

Could you take another look when you get a chance?

0hmX:
@itisrohit Thanks for working on this, but I’m not seeing any improvements yet. We consider something an improvement if it visibly fixes something, like a broken board, or even shows benchmark improvements.

So right now, it’s hard for me to tell whether it’s actually working or not. Thanks again for working on this. You could try adding more explicitly broken repos and show that they are being fixed in this PR.

You can also let me know if I missed any snapshots that you think prove this fixes a problem.

PR #1246, opened 2026-05-23
Limit Pipeline 7 component regions to BGA-like components

ShiboSoftwareDev:
before:
image
after:
image

Pipeline 7 was treating every componentId as a component region, which caused ordinary passives/connectors in circuit-json-derived SRJ to produce thousands of extra topology regions. This narrows component detection to BGA-like dense pad grids, keeps non-BGA component pads in the global topology, and adds a USB-C repro asserting the topology no longer creates non-BGA component regions.


tests/repro/usb-c-power-adapter-pipeline7.test.ts
```diff
@@ -0,0 +1,207 @@
...
+  return null
+}
+
+const getObstacleLayers = (element: CircuitJsonElement) => {
+  if ("layers" in element) return element.layers
+  if ("layer" in element) return [element.layer]
+  return ["top"]
+}
+
+const convertCircuitJsonToSimpleRouteJson = (
```
seveibar: slightly concerning

seveibar (approved):

PR #1255, opened 2026-05-23
Preserve multilayer cramped ports for OV5640 routing

ShiboSoftwareDev:
Keeps cramped segment ports that touch multilayer regions so dense component escapes remain routable after cramped-port filtering. Adds an OV5640 pipeline 7 repro fixture and final-solve SVG snapshot, and updates the tiny-hypergraph dependency needed for the route to complete.


seveibar:
@ShiboSoftwareDev introduce portPenalty for cramped port points, the relaxed DRC score will go way up

seveibar:
e.g. for the duplicated port points, i set it to 1000, so they're only used as a last resort

0hmX:
/benchmark --pipeline 7 --dataset 18 --effort 10

PR #1263, opened 2026-05-24
Allow via-sized side cells in BGA topology

seveibar:


seveibar:
/benchmark --pipeline 7 --dataset 18 --effort 10

PR #1298, opened 2026-05-26
add prev and next portPointId

0hmX:
- **add prev and next  port point ids**
- **svg update**


seveibar:
small gain on drc, nice

PR #1304, opened 2026-05-27
Cap the minTraceWidth to the terminal pads dimension

imrishabh18:


seveibar (commented):
i don't think this is the right approach- we essentially only care about the pad width when we're _close to the pad_, we don't want to penalize the entire route

seveibar:
nice work generally tho (CC @Sang-it )

PR #1309, opened 2026-05-28
Converge traces which have traceThickness more than the terminal pad dimension

imrishabh18:
| Before | After |
|--------|-------|
| image | image |


seveibar (approved):

seveibar:
Trace width solver should maybe be a pipeline

PR #1312, opened 2026-05-29
move to Base Pipeline Solver and automatic marking of step number in graphics object

0hmX:


seveibar (changes requested):
Messy? Reminds me of enterprise codebases where they’re always wrapping things instead of doing proper refactors untila codebase is unusuable

PR #1313, opened 2026-05-29
avoid exploding topology size from BGAs

seveibar:


seveibar:
/benchmark --pipeline 7 --dataset 18 —sample-timeout 600s

0hmX:

0hmX (approved):

seveibar:
@0hmX i wonder if we should attempt to fix partial bgas instead of this hack tho :/

PR #1327, opened 2026-06-02
Implement trace thickness as a parameter

JirA44:
Here's a concise pull request description:

Implemented support for trace thickness multiples in the `Trace` and `Router` classes as per repo code. This fix allows for more precise control over the thickness of traces, which should improve accuracy in simulations. The change was made by introducing a new `TraceThickness` enum with a `SingleFoil` variant that takes an optional thickness multiplier, and updated the `Trace` struct to include this field. You can verify this fix by running benchmarks with different trace thickness settings.

Closes #66


JirA44:
Closing to resubmit with a proper TypeScript implementation.

PR #1328, opened 2026-06-02
Implement trace thickness as a parameter

JirA44:
This pull request addresses issue #66 by introducing a `TraceThickness` class that allows for customizable trace thickness in our benchmarks. Instead of hardcoding specific values, we've made the 2x, 4x, and 8x multiples configurable. To verify this fix, you can run your benchmarks as usual to see if the updated workflows are producing expected results, or check the new instructions in `benchmark-instructions.yml` for any references to trace thickness configurations.

Closes #66

### DIFF
--- a/implement_trace_thickness_as_a_para.py
+++ b/implement_trace_thickness_as_a_para.py
@@ -1,0 +1,27 @@
+import math
+
+class TraceThickness:
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


JirA44:
Closing to resubmit with a proper TypeScript implementation.

PR #1331, opened 2026-06-03
repro: document Pipeline8 preplaced-via layer-span validation gap

itisrohit:
### Summary
Adds a focused stitch-solver repro test for Pipeline8 preplaced-via validation.

### Bug
The current preplaced-via allowlist is keyed only by XY coordinate.

That means two constrained preplaced vias at the same XY collapse into a single allowed point. The stitch solver can then permit a direct layer transition that is not actually supported by the via stack.

### What the test shows
- a blocked solver correctly fails with:
  
- the current XY-only allowlist then bypasses that failure for the same geometry

### DIFF
--- a/tests/stitch-solver/single-high-density-route-stitch-preplaced-via-layer-span.test.ts
+++ b/tests/stitch-solver/single-high-density-route-stitch-preplaced-via-layer-span.test.ts
@@ -1,0 +1,52 @@
+import { expect, test } from "bun:test"
+import { SingleHighDensityRouteStitchSolver3 } from "lib/solvers/RouteStitchingSolver/SingleHighDensityRouteStitchSolver3"
+import { getAssignableViaPointKeys } from "lib/autorouter-pipelines/AutoroutingPipeline8/assignableViaUtils"
+import type { Obstacle } from "lib/types"
+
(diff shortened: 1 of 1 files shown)


seveibar (commented):
can you add a visual snapshot, i don't really understand what's happening- is there a bug report relevant to this?

seveibar (approved):
CC @0hmX

itisrohit:
Closing this repro because I could not validate a real board-level violation for the exact case this test models.

What I did check:
- Pipeline8 is still reachable: `core` selects `AutoroutingPipelineSolver8` for `useLaserPrefabSolver`, so this is not purely dead code.
- `netIsAssignable` is still part of the public `laser_prefab` story.
- Existing local Pipeline8 fixtures and public assignable-via SRJ dataset cases did not produce an unsupported via layer-span output for this same-XY partial-stack scenario.

So the current test demonstrates an internal invariant mismatch in the XY-only allowlist, but I do not have enough evidence that it currently causes a bad routed board for users in production.

I would rather close this repro than overclaim impact. If we find a concrete routed board where Pipeline8 / `laser_prefab` emits an invalid layer transition through prefabricated vias, we should reopen with that board-level violation.

PR #1335, opened 2026-06-03
Refactor component detection into detector modules

seveibar:
## Summary
- Split component classification into per-type detector modules under `detectors/` for BGA, SOIC, QFP, and QFP thermal-pad.
- Added a shared detector interface and registry so `memberObstacles` can be evaluated through a common `createComponentDetector` entrypoint.
- Kept the existing detection order and stage behavior intact while moving the shape-specific logic out of the solver stage.

## Testing
- Ran the focused component-detection test file successfully.
- Ran TypeScript `noEmit` validation successfully against the local toolchain.


0hmX:
lgtm

PR #1343, opened 2026-06-05
move componentDetectionSolver to just after preprocessSimpleRouteJsonSolver

0hmX:


lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -267,11 +267,16 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
...
+      "componentDetectionSolver",
+      ComponentDetectionSolver,
+      (cms) => [{ inputSrj: cms.srj }],
+    ),
     definePipelineStep(
       "escapeViaLocationSolver",
       EscapeViaLocationSolver,
       (cms) => [
-        cms.srj,
+        cms.componentDetectionSolver!.getOutput().global,
```
seveibar: that's a bit unclear, `global = globalSrj`?

seveibar (approved):

0hmX:
While trying to rename this, I realized it could cause problems in the future.

Basically, this SRJ was acting as an input only for the topology solver. It removes some of the connections, so any connection that goes inside the component gets removed, and the component itself is replaced with an obstacle.

This was not a problem before because, after merging the topology, we used the original SRJ, which  still had everything. This modified SRJ should probably have its own type, and that type should be the input for the topology solver. Right now, everything takes an SRJ as input, so this was a bit of a hack.

Now that we are moving this to the second stage, the remaining connections are being used to create vias for escape routing. It is fine for now, but I think in the future we might run into problems because of this SRJ object with missing connections.

its fine for now

seveibar (approved):

0hmX:

0hmX:

PR #1345, opened 2026-06-07
Add SRJ19 BGA passive-overlay dataset to benchmark runner and fixtures

Abse2001:


Abse2001:
/benchmark --pipeline 7 --dataset 19 --sample-timeout 600s

PR #1346, opened 2026-06-07
repro: SameNetViaMergerSolver leaves a route-geometry stub when consolidating overlapping same-net vias

itisrohit:
### Summary
Adds a deterministic reproduction test in `tests/repro/same-net-via-merger-phantom-copper-stub.test.ts` documenting a **correctness/cleanliness** issue in `SameNetViaMergerSolver`: when consolidating overlapping same-net vias, it leaves a there-and-back "stub" in the route geometry.

**This is not an electrical/DRC hazard** (see geometry note below) — it is a correctness issue. The test asserts the *current* behavior so CI stays green ("no failing tests on merge"); no fix is included.

### What happens
When two same-net vias overlap, the merger correctly consolidates them to a single via. But `moveViaTo` (`lib/solvers/SameNetViaMergerSolver/SameNetViaMergerSolver.ts`) does this by *inserting* a detour to the kept via's location into the route geometry **without removing the original via point**. Result: the route runs from the kept location out to the removed via's old location and back, on both layers.

Two concrete inconsistencies this leaves behind:
- The route geometry no longer matches the intended route (an extra there-and-back excursion that was never routed).
- `route.vias` (consolidated, looks clean) disagrees with the actual copper in `route.route` — a footgun for any downstream code that trusts route geometry.

### Severity / scope (not a hazard)
The stub runs between the two via centers. Vias only merge when their pads overlap, so the stub sits **inside the merged via-pad footprint** for essentially the whole overlap range — it only pokes out by ~0.04–0.06 mm in a razor-thin band where vias are barely touching, which is within fab tolerance. It is same-net copper, so **no net-to-net clearance/electrical effect**. This is a correctness/cleanliness bug, not a DRC issue.

### To run
```
bun test tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
```

### DIFF
--- a/tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
+++ b/tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
@@ -11,9 +11,15 @@
  *
  * This stub is NOT cleaned up by the downstream path_simplification stage — it
  * survives into the final routed output (verified separately via
- * TraceSimplificationSolver). It is same-net copper, so it is not a net-to-net
- * short, but it is spurious/dangling copper (stub → antenna/reflection,
- * manufacturing) and means the output geometry does not match the intended route.
+ * TraceSimplificationSolver).
+ *
+ * This is a correctness/cleanliness issue, NOT an electrical/DRC hazard. The stub
+ * runs between the two via centers, and vias only merge when their pads overlap,
+ * so the stub sits inside the merged via-pad footprint (it only pokes out ~0.04-
+ * 0.06mm in a razor-thin band where vias are barely touching — within fab
+ * tolerance). It is same-net copper, so there is no net-to-net clearance effect.
+ * The defects are: (1) the route geometry no longer matches the intended route,
+ * and (2) route.vias disagrees with the actual copper in route.route.
  *
  * Per "no failing tests on merge", this test asserts the CURRENT (buggy) behavior:
  * the merge ADDS same-layer copper that did not exist in the input. A merge that
--- a/tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
+++ b/tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
@@ -1,0 +1,112 @@
+/**
+ * Repro: SameNetViaMergerSolver leaves a phantom copper "stub" when consolidating
+ * overlapping same-net vias.
+ *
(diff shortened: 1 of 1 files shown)


tests/repro/same-net-via-merger-phantom-copper-stub.test.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * Repro: SameNetViaMergerSolver leaves a phantom copper "stub" when consolidating
+ * overlapping same-net vias.
+ *
+ * When two same-net vias overlap, the merger relocates one via onto the other
```
0hmX, line 5: Good work on the repo.

But is this really a problem? Yes, it is a bit misleading, but I don’t think it is technically wrong because it is the same net, so it does not really matter.

If you can find an actual bad downstream effect caused by this, then I think we should take a look at it. But right now, this does not look like an issue or something that needs to be fixed.

0hmX (changes requested):
Cool you are working you need to add some svg snapshots

itisrohit:
Fair point, I went and checked the geometry to see if there's a real downstream effect, and you're right.

The stub always runs between the two via centers, and two vias only merge when their pads overlap (centers closer than a via diameter). Whenever they overlap, the two circular pads already cover that whole segment, so the stub sits *inside* the merged via-pad footprint:

| via separation | stub vs. merged via-pad footprint |
|---|---|
| 0.10 mm | 0.066 mm **inside** the pads |
| 0.20 mm | 0.037 mm inside |
| 0.25 mm | 0.008 mm inside |
| 0.29 mm | +0.037 mm sticks out |
| 0.299 mm | +0.063 mm sticks out |

It only pokes out at all in a razor-thin band where the vias are *barely* touching (~0.29–0.30 mm apart), and even then by ~0.04–0.06 mm — within fab tolerance, and a different net would have to be illegally close to the pads to be affected. So no realistic net-to-net clearance / electrical effect. Same-net, as you said.

So I'll reframe this: it's a **correctness/cleanliness** issue, not a hazard — the route geometry no longer matches the intended route, and `route.vias` disagrees with the actual copper geometry (a footgun for anything downstream that trusts route geometry). Not an electrical defect.

Happy to leave the repro as documentation of that geometry-vs-vias-list inconsistency, or close it, your call.

PR #1357, opened 2026-06-08
Fix Pipeline 7 hang on invalid topology node dimensions

ShiboSoftwareDev:
Prunes non-finite capacity mesh node dimensions before node subdivision so the aspect-ratio loop cannot run forever. Adds regression coverage for Infinity and NaN node dimensions.


0hmX (approved):
I think it’s fine for now.

But the fact that this is even happening is bad. It means something went wrong, possibly due to invalid data or division by NaN. I’m not sure yet.

This needs more investigation.

PR #1362, opened 2026-06-09
Fix high-density solver metadata for child solvers without getSolverName

ShiboSoftwareDev:


lib/solvers/HighDensitySolver/HighDensitySolver.ts
```diff
@@ -210,7 +210,27 @@ export class HighDensitySolver extends BaseSolver {
       return this.getIntraNodeStrategyName(solver.hyperParameters)
     }
 
-    return solver.getSolverName()
+    if (
+      solver &&
+      typeof solver === "object" &&
+      "getSolverName" in solver &&
+      typeof solver.getSolverName === "function"
```
0hmX: this seems like an hack is a better solution not possible?

0hmX (approved):

lib/solvers/HighDensitySolver/HighDensitySolver.ts
0hmX: maybe let the ai update all the sub solver so no one is missing getSolverName

PR #1367, opened 2026-06-10
fix(autorouter): same-net via clearance + per-pour escape grouping

gwhthompson:
## Summary

Two `EscapeViaLocationSolver` fixes for pads the author has already stitched to a copper pour with their own vias (e.g. GND pads stitched to an inner plane on a 4-layer board):

- **Same-net vias no longer block escape-via placement.** Both clearance checks (`hasClearEscapePath`, `getMinBlockingClearance`) treated every overlapping obstacle as a blocker — including a same-net stitch via sitting exactly where the escape via should land — so a pad ringed by its own stitch vias could get no escape via at all. They now share an `isSameNetObstacle` predicate that skips non-pour obstacles on the connection's own net (the span check additionally requires the obstacle to cover the full source→target layer pair, since it measures clearance across that span). The skip only ever *relaxes* clearance for the obstacle's own net, so no previously-valid candidate is rejected.
- **One contiguous pour now produces one escape-via target.** Rasterised pours were keyed per obstacle cell (`getObstacleKey`), so escape vias landing in different cells of the same pour were never grouped, and the router added a redundant pad-to-pad surface trace between pads that are already connected through the pour. Pours are now keyed by `(layer, net)` via a `getPourKey` helper; `getObstacleKey` had no other caller and is removed.

## Testing

- Two regression tests, each of which fails on the pre-change solver:
  - `pour-via-escape-same-net-via-clearance`: a pad ringed by same-net spanning vias still gets an escape via (without the skip, every candidate is blocked and no via is placed); a foreign-net ring still blocks (negative control).
  - `pour-via-escape-pour-key-collapse`: two pads over adjacent cells of one contiguous pour group into a single escape-via target (the old per-cell key left them ungrouped and the router routed pad-to-pad).
- Full `tests/features/pour-via-escape/` suite green; the existing `01–06` escape-via snapshots are unchanged.
- `bunx tsc --noEmit` clean; `biome format` clean.
- Benchmark (`./benchmark.sh AutoroutingPipelineSolver4 40 --dataset dataset01`): completion 92.5%, relaxed-DRC 87.5%, 0/40 timeouts, P50 2.2 s, P95 9.9 s — unchanged vs `main`, identical failure samples (all pre-existing). The solver only acts on nets that have a matching copper pour and no shipped dataset contains one, so the change is inert on the benchmark; the regression tests above carry the behavioural proof.

### DIFF
--- a/lib/solvers/EscapeViaLocationSolver/EscapeViaLocationSolver.ts
+++ b/lib/solvers/EscapeViaLocationSolver/EscapeViaLocationSolver.ts
@@ -78,7 +78,10 @@ const pointMatches = (
 
 // A pour is rasterised into many obstacle cells sharing the same layer + net;
 // key by (layer, net) so every cell of one pour groups into a single escape-via
-// target instead of one target per raster cell.
+// target instead of one target per raster cell. This assumes one contiguous
+// pour per (layer, net): two physically-disjoint same-net regions on one layer
+// are intentionally not split — supported boards pour one contiguous region per
+// (layer, net).
 const getPourKey = (pour: Obstacle) =>
   `${pour.layers[0]}:${[...new Set(pour.connectedTo)].sort().join(",")}`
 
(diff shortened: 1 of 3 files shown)


gwhthompson:
Same as my core one (tscircuit/core#2416): stale bot closed this for inactivity too. Still mergeable, no conflicts with main. It fixes escape vias for pads already stitched to a plane: same-net stitch vias no longer block the pad's own escape via, and each pour collapses to one escape target per (layer, net) instead of one per raster cell. Could a maintainer reopen? Read-only here so I can't do it myself, can rebase or take feedback.

PR #1393, opened 2026-06-12
feater: enable sparse candidate storage in tiny hypergraph

0hmX:


seveibar (approved):
i think it's important to remove unused code as well, i think you reverted the other sparse data structure but also removing this flag and just making sparse default etc.

PR #1408, opened 2026-06-18
preprocessSimpleRouteJsonSolver: snap point to connect to nearest pad for rotated rect cases where the point is outside and enables STATIC_REACHABILITY_PRECHECK in DuplicateCongestedPortSolver

0hmX:
origin
image

after running preprocessSimpleRouteJsonSolver

before fix:
image

after fix:
image

### DIFF
--- a/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
+++ b/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
@@ -758,7 +758,7 @@ export class TinyHypergraphPortPointPathingSolver extends BaseSolver {
               2_000_000 * getEffortScale(params.effort),
             ),
             RIP_THRESHOLD_RAMP_ATTEMPTS: 0,
-            STATIC_REACHABILITY_PRECHECK: false,
+            STATIC_REACHABILITY_PRECHECK: true,
           },
         },
       )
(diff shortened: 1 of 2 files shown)


lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
```diff
@@ -758,7 +758,7 @@ export class TinyHypergraphPortPointPathingSolver extends BaseSolver {
               2_000_000 * getEffortScale(params.effort),
             ),
             RIP_THRESHOLD_RAMP_ATTEMPTS: 0,
-            STATIC_REACHABILITY_PRECHECK: false,
+            STATIC_REACHABILITY_PRECHECK: true,
```
0hmX, line 761: This could have been a separate PR, but `TinyHypergraphPortPointPathingSolver` was getting stuck at the construction level for a really long time because `DuplicateCongestedPortSolver` could not early-detect a case where no path was possible.

DuplicateCongestedPortSolver run inside the constructor with DuplicateCongestedPortSolver.solve()

I only found out because I did a refactor that is too large, but it made `DuplicateCongestedPortSolver` **a proper solver with visualization**.

It is too large, and a smaller version is possible https://github.com/tscircuit/tscircuit-autorouter/pull/1407 its in draft

you can see the vercel preview go to srj18 problem 004 its getting stuck but now with visualization

The start mesh node was disconnected from the world mesh, but it still did not exit early and kept running A*. So we need this to be on, so it stops early. We were spending like 10K+ steps, which is an insane number and caused my browser to hang.

Basically this flag help skip that issue of disconnected mesh not being detected

seveibar (approved):

PR #1411, opened 2026-06-18
add tolerance to the distance check

0hmX:
image

Even though the points to connect were inside, it was causing the solver to fail. Add a tolerance, because before we were using `== 0`, which was too strict.


0hmX:
/benchmark --dataset 18 --pipeline 7  --sample-timeout 1200s

0hmX:
/benchmark --dataset 18 --pipeline 7 --sample-timeout 300s

PR #1423, opened 2026-06-19
update dataset srj 18

AnasSarkiz:
add missing vias


AnasSarkiz:
/benchmark --dataset dataset18 --pipeline 7 --timeout 600s

AnasSarkiz:
/benchmark --dataset 18 --pipeline 7 --sample-timeout 1200s

0hmX (approved):
The worst part is that, because we now have gaps, we also have these tiny rectangles that overlap.

It is all valid though. Overlapping rectangles are allowed; it is just that said rectdiff algo could not handle this case, mostly because the rectangles are too small.


Image

PR #1438, opened 2026-06-24
Fix publish workflow auto-merge fallback

0hmX:
## Problem

The `Publish to npm` workflow could fail after the package was already published. The failing step was `Enable auto-merge`, where this command ran against the generated version-bump PR:

```bash
gh pr merge <pr> --auto --squash --delete-branch
```

GitHub sometimes creates that PR in a state where all required checks are already satisfied, so there is nothing for auto-merge to wait on. In real-life terms: the workflow was asking GitHub to "merge this later when it is ready", but GitHub was saying "it is ready right now, so there is no later auto-merge to enable."

The observed error was:

```text
GraphQL: Pull request Pull request is in clean status (enablePullRequestAutoMerge)
```

That left version-bump PRs open and stale even though the release itself had completed.

## Fix

The workflow still tries auto-merge first. If GitHub returns the known clean-status `enablePullRequestAutoMerge` error, it falls back to a direct squash merge. Other GitHub CLI errors still fail the job.

## Validation

- Parsed the workflow YAML.
- Checked the edited shell block with `bash -n`.
- Simulated the clean-status fallback and unrelated-error paths.
- Ran `bun run format`; it failed on existing `.vendor/rectdiff` Biome config incompatibilities and reported no fixes applied.

### DIFF
--- a/.github/workflows/bun-pver-release.yml
+++ b/.github/workflows/bun-pver-release.yml
@@ -54,8 +54,22 @@ jobs:
 
       - name: Enable auto-merge
         if: steps.create-pr.outputs.pull-request-number != ''
+        shell: bash
         run: |
-          gh pr merge ${{ steps.create-pr.outputs.pull-request-number }} --auto --squash --delete-branch
+          PULL_REQUEST_NUMBER="${{ steps.create-pr.outputs.pull-request-number }}"
+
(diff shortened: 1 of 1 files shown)


seveibar (approved):
you need to track if this is ever run, because it's much more confusing than the original and i don't think it will pass the permissions setup (because automerging was a workaround for permissions issues to begin with iirc)

seveibar:
(so you can merge but revert if it doesn't work)

PR #1505, opened 2026-07-05
Fix False Source Trace Contiguity Errors by Separating Obstacle Connectivity from Trace Endpoints

AnasSarkiz:
### Summary

Fixes a source-trace conversion bug where obstacle-derived same-net connectivity was being treated as required physical trace endpoints.

The converter now keeps actual route endpoints in `connected_source_port_ids` and moves obstacle-derived connectivity into `connected_source_net_ids`. This preserves same-net clearance behavior while preventing `checkTracesAreContiguous` from reporting false missing-pad connections.

### Problem

`convertToCircuitJson` previously appended obstacle `connectedTo` metadata into `source_trace.connected_source_port_ids`.

That made DRC contiguity checks overreach:

- `connected_source_port_ids` is interpreted as the list of physical ports a trace must connect to.
- Obstacle `connectedTo` can include same-net ports/pads that are electrically related but are not endpoints of this routed trace.
- As a result, valid routes were reported as missing connections to unrelated same-net pads.

This showed up in `dataset01` as false `pcb_trace_error` failures, especially in `circuit003`.

Screenshot 2026-07-05 at 4 16 43 PM

Screenshot 2026-07-05 at 4 17 44 PM


lib/testing/utils/convertToCircuitJson.ts
```diff
@@ -208,6 +208,12 @@ function convertHdRouteToCircuitJsonTraces(
   return traces
 }
 
+const getObstacleConnectivityIds = (obstacles: Obstacle[]) =>
+  obstacles.flatMap((obstacle) => [
+    `obstacle_${obstacle.center.x.toFixed(3)}_${obstacle.center.y.toFixed(3)}_${obstacle.layers.join(".")}`,
```
seveibar: Why add the made up net?

seveibar (commented):
Seems ok to me but gonna let ankan review!

lib/testing/utils/convertToCircuitJson.ts
AnasSarkiz: Now `connected_source_net_ids = real same-net IDs from obstacle.connectedTo`

tests/features/pipeline7-dataset01-source-trace-contiguity.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      "pcb_port_28",
+    ])
+    expect(sourceTrace19?.connected_source_net_ids).toContain("pcb_port_24")
+    expect(sourceTrace19?.connected_source_net_ids).toContain("pcb_port_26")
+
+    const { errors } = getDrcErrors(circuitJson, RELAXED_DRC_OPTIONS)
+
+    expect(errors).toHaveLength(0)
+  },
+  { timeout: 60_000 },
```
0hmX: remove the timeout, its generally set using CI

tests/utils/convertToCircuitJson-obstacle-port-attribution.test.ts
```diff
@@ -187,3 +188,113 @@ test("uses original SRJ obstacle rotation when creating circuit-json pads", () =
     ccw_rotation: 45,
   })
 })
+
+test("keeps obstacle connectivity out of source trace endpoint expectations", () => {
```
0hmX: this should move into its own file one test per file rule

0hmX (approved):
lgmt

Issue #1507, opened 2026-07-06
Same-layer crossings can survive to output as net shorts (crossings modeled as via-resolvable, not guaranteed resolved)


maci0:
## Summary

On dense boards the pipeline can emit **same-layer crossings between different nets** — i.e. real shorts in the routed output (`getDrcErrors` flags them, same class as `tests/bugs/simplification-trace-overlap-repro`).

## Root cause

`lib/solvers/UnravelSolver/calculateCrossingProbabilityOfFailure.ts` treats a same-layer crossing as **via-resolvable capacity cost**, not a hard constraint:

```ts
const estNumVias =
  numSameLayerCrossings * 0.82 +
  numEntryExitLayerChanges * 0.41 +
  numTransitionCrossings * 0.2
const estUsedCapacity = (estNumVias / 2) ** 1.1
// probability of failure = estUsedCapacity / totalCapacity
```

(only the 1-layer case returns a hard `1`). So the topology stage
(`computeCostPerRegion`) accepts a plan with same-layer crossings whenever the
node still has capacity, on the assumption the crossing will later be resolved
by a via / layer change. When the downstream HighDensity + `HighDensityRepairSolver`
stage **cannot** actually resolve it (no room for the via in a tight node), the
crossing survives into the final geometry as a different-net same-layer overlap =
short. Nothing guarantees the output is same-layer-crossing-free.

## Concrete case

A dense 2-layer board: nRF52840 module (E73, 1.27mm castellated edge + interior
LGA) next to a USB-C receptacle pad row. The autorouter shorted `USB_DM <-> VBUS`
and left a `tracks_crossing`. Moving the offending signals off the module's
**interior LGA pads onto edge castellations** (clean escape) made the specific
crossings disappear — consistent with the cost model committing to
interior-pad topologies whose crossings the HD stage then can't resolve.

## Suggested direction

- A final guarantee pass: after HighDensity/repair, detect any residual
  different-net same-layer overlap (`getDrcErrors`) and either rip+reroute that
  region or fail the node, rather than emitting a short; **or**
- In `calculateNodeProbabilityOfFailure`, weight `numSameLayerCrossings` by
  whether the node actually has via capacity to resolve it (an unresolvable
  crossing should approach a hard `1`, not a soft capacity fraction).

Happy to help with a minimal SimpleRouteJson fixture + a `getDrcErrors`-based
regression test if that's useful. Related theme (schematic side): tscircuit/core#2552.

[...]

PR #1513, opened 2026-07-07
Guarantee no different-net same-layer shorts in routed output (#1507)

maci0:
Fixes the root of #1507: the router can emit different-net same-layer overlaps (real shorts) / sub-clearance near-misses, because same-layer crossings are modeled as via-resolvable cost and **nothing downstream validates the output**. Which boards come out shorted is effectively non-deterministic.

I first tried the tractable route (bumping the crossing weight in `calculateNodeProbabilityOfFailure`, `0.82` -> a principled `1.0`) and measured it against the bug-report fixtures with `getDrcErrors`: **8 fixtures 11 -> 7, but 16 fixtures 37 -> 65 (regression).** A cost-tune just shifts which fixtures short, so it's not the fix.

## This PR
`guaranteeNoSameLayerShorts(traces, clearance)` — a pure pass over `SimplifiedPcbTraces`: if two different-connection traces run under `clearance` on the same copper layer (0 gap = a hard short), truncate one at the offending segment so it degrades to an **unrouted ratsnest (recoverable)** rather than a **short (broken board)**. Output is then guaranteed free of different-net same-layer violations, independent of any topology-cost tuning. Reuses `minimumDistanceBetweenSegments`; `connection_name` is net-attached so same-net traces are skipped.

## Validation
- Unit: two crossing traces -> one truncated, the non-crossing net untouched.
- Integration: `bugreport01` — raw solver output carries a different-net same-layer violation; the pass removes it (0 after).

## Notes for review
- Additive only (one util + one test); no existing code changed, so no snapshots move.
- Suggested wiring: the tail of each pipeline's `getOutputSimplifiedPcbTraces`. There's no shared base, and wiring flips the routed-vs-unrouted contract on genuinely-shorting boards, so I left the wiring to you rather than touching all ~10 pipelines + their snapshots.
- Follow-up: resolve the crossing with an **inserted via** (keep the net routed) instead of truncating.

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
@@ -60,6 +60,7 @@ import { StrawSolver } from "../../solvers/StrawSolver/StrawSolver"
 import { TraceSimplificationSolver } from "../../solvers/TraceSimplificationSolver/TraceSimplificationSolver"
 import { TraceWidthSolver } from "../../solvers/TraceWidthSolver/TraceWidthSolver"
 import { PreprocessSimpleRouteJsonSolver } from "../AutoroutingPipeline4_TinyHypergraph/PreprocessSimpleRouteJsonSolver"
+import { guaranteeNoSameLayerShorts } from "lib/utils/guaranteeNoSameLayerShorts"
 
 interface CapacityMeshSolverOptions {
   capacityDepth?: number
@@ -978,7 +979,11 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
       }
     }
 
-    return traces
+    // Safety net: a router must never emit a short. The capacity pipeline models
+    // same-layer crossings as via-resolvable cost, so an unresolved one can
+    // survive here as a different-net same-layer overlap. Truncate any such
+    // offender to a ratsnest (recoverable) rather than ship a shorted board.
+    return guaranteeNoSameLayerShorts(traces, 0)
   }
 
   getOutputSimpleRouteJson(): SimpleRouteJson {
--- a/lib/utils/guaranteeNoSameLayerShorts.ts
+++ b/lib/utils/guaranteeNoSameLayerShorts.ts
@@ -1,0 +1,104 @@
+import { minimumDistanceBetweenSegments } from "lib/utils/minimumDistanceBetweenSegments"
+import type { SimplifiedPcbTraces } from "lib/types/srj-types"
+
+/**
+ * A router must never emit a short. The capacity pipeline models same-layer
+ * crossings as via-resolvable cost, so when the high-density/repair stage can't
+ * actually resolve one it survives into the output as a different-net same-layer
+ * overlap (a real short) or a sub-clearance near-miss. Nothing downstream checks
+ * for this, so which fixtures come out shorted is effectively non-deterministic.
+ *
+ * This is the safety net: after routing, if two traces of different connections
+ * run closer than `clearance` on the same copper layer (0 gap = a hard short),
+ * truncate one at the offending segment so it degrades to an unrouted ratsnest
+ * (recoverable) instead of a short / DRC-fail (a broken board). The output is
+ * then guaranteed free of different-net same-layer violations, independent of any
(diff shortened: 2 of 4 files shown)


seveibar:
@maci0 can you share the board or a bug report? We want to understand the motivation more- generally to keep the codebase simple we try to understand each fix pretty deeply/visually- this is usually easiest with a reproduction

maci0:
@seveibar here's a minimal, deterministic reproduction (pushed as `tests/bugs/same-layer-short-crossbar-repro.test.ts`).

It's a crossbar: 22 nets whose endpoints are reversed left↔right, packed into a 2 × ~8.4 mm window on 2 layers, so every net must cross every other. That saturates the via budget and forces unresolved same-layer crossings into the routed output.

Checking the routed output (`mergedHdRoutes` → `getDrcErrors`), this produces **26 different-net same-layer overlaps** — real shorts ("overlaps with trace"). The count is a smooth function of density, not a magic fixture: n=8 → 0, n=18 → 1, n=22 → 26. That matches the "which fixtures come out shorted is effectively non-deterministic" framing in #1507 — it's a via-budget threshold.

Root cause is as #1507 describes: `calculateCrossingProbabilityOfFailure` treats a same-layer crossing as via-resolvable *cost*, so when the HD/repair stage can't actually resolve one, it survives to output.

One thing I noticed wiring this up: `guaranteeNoSameLayerShorts` isn't called anywhere in the pipeline yet — it's defined but never applied to `getOutputSimplifiedPcbTraces`, so the routed output is still shorted. Want me to wire it into the output path so this repro asserts 0 (and add the before/after to the test)?

maci0:
Wired it up in 81b6facc: `guaranteeNoSameLayerShorts` now runs in `AutoroutingPipelineSolver7_MultiGraph.getOutputSimplifiedPcbTraces()` (clearance 0 = hard overlaps). The repro is now a before/after — the raw merged routes still short (26), the pipeline output is 0. The related overlap tests still pass.

As noted, this is a safety net: it truncates the offender to a ratsnest, trading a short for a recoverable open. The root fix (make a same-layer crossing a hard constraint in `calculateCrossingProbabilityOfFailure`, or resolve it with an inserted via) is the follow-up flagged in the file — happy to go that route instead if you'd prefer.

maci0:
@seveibar here's a standalone version you can run straight against the published package, no need to check out this branch or touch any internals ([gist](https://gist.github.com/maci0/a460424fa5215d5da37771144685c91d)):

```
npm i @tscircuit/capacity-autorouter@0.0.654 && node repro-1513.mjs
```

It's fully synthetic, no real board: N nets with endpoints reversed left<->right (so every net wants to cross every other), squeezed beside a row of bottom-layer copper keep-out that starves the via budget, a generic stand-in for a castellated module sitting next to a connector pad row. Then it reads the router's **public** output (`getOutputSimpleRouteJson()`) and checks for different-net segments that properly cross on the same copper layer.

On 0.0.654 the routed output carries real different-net same-layer crossings (shorts):

```
n=8:  solved=true traces=8  different-net same-layer crossings=1
n=10: solved=true traces=10 different-net same-layer crossings=1
n=12: solved=true traces=12 different-net same-layer crossings=7
    n=8 example: nets c7 x c1 cross on layer "top"
```

Deterministic run-to-run (fixed shuffle seed), and the count climbs with density (1 -> 1 -> 7). That's the shape of #1507: it's a via-budget threshold, so *which* board comes out shorted looks non-deterministic even though a given board is stable.

The real motivating case (I can't share the board, it's private, but in words): a dense 2-layer ~24x56 mm board with an E73 nRF52840 castellated module sitting directly beside the USB-C receptacle pad row. The router intermittently emits a VBUS<->GND (and earlier USB_DM<->VBUS) same-layer crossing that KiCad DRC flags, and it reroutes away non-deterministically seed to seed. Same failure the synthetic board reproduces.

On your "keep it simple" point, I agree, and I don't want to push weight the codebase doesn't need. Two things that hopefully make this palatable: it's additive and currently unwired (one util + one test), so it moves no snapshots; and where it does fire it degrades a short to an unrouted ratsnest (recoverable) rather than shipping a broken board. The nicer long-term fix is to actually resolve the crossing with an inserted via so the net stays routed, instead of truncating. Happy to pursue that follow-up (or drop this entirely) if you'd prefer that over a truncating safety net.

maci0:
Reopened as a fresh PR against current main: https://github.com/tscircuit/tscircuit-autorouter/pull/2553 (stale-bot closed this one).

PR #1530, opened 2026-07-07
fix: connect same-region connections (direct segment for degenerate HD route)

### LINKED ISSUE #1528: Trace left as disconnected same-layer fragments near a pad (net not stitched to completion)
On a real board, one net (V3V3 -> U1 pin 6 VDDIO) comes out of the router as **two disconnected F.Cu segments that never join**, leaving the pad unconnected:

```
Track [V3V3] on F.Cu, length 0.2436 mm @ (94.84, 94.80)
Track [V3V3] on F.Cu, length 0.9956 mm @ (95.08, 95.60)
Pad 6 [V3V3] of U1 on F.Cu     @ (95.08, 94.80)
```

- All three are on the **same layer** (F.Cu), roughly collinear (x≈95.08), with a **clear path** between them (no pad or obstacle in the ~0.8mm gap - verified by scanning all footprint pads in the region).
- The router produced the trace *fragments* but did not stitch them to each other or to the pad.
- **Effort-independent**: identical result at effort 1, 3, 10, 25 (the board otherwise routes in ~7s, so this isn't an iteration-budget issue). It's deterministic.
- Every other net on the board (21/22) routes and connects cleanly.

Looks like a completeness gap in the final route-stitching / pad-connection step of the TinyHypergraph pipeline: it can leave a net as disjoint same-layer segments instead of one connected route (or failing the net outright).

Repro board: nRF52840 module + BME690 (U1) + OLED + USB-C, dual GND `<copperpour>`; available on request. Happens on `AutoroutingPipelineSolver4` (the default).

maci0:
## Problem
When both endpoints of a connection fall inside a **single capacity node** (e.g. two adjacent same-net pads), no region boundary is crossed. `buildInputNodesWithPortPoints` filters out the `_tinyTerminal` port points (they are normally attached only at route ends), so the HD solver has nothing to route and returns a **degenerate single-point** `hdRoute`. The net is then left as a floating stub, unconnected.

Observed on a real board: net V3V3 -> U1 pin6 came out as two disconnected same-layer fragments; the intra-node edge between adjacent pins 5/6 (both on V3V3) never got a trace. Deterministic, effort-independent. (#1528)

## Fix
`convertHdRouteToSimplifiedRoute` already receives the connection's `connectionPoints`. When the produced route is degenerate (< 2 distinct wire points) and there are two distinct endpoints, emit the **direct segment** between them (same layer -> one wire pair; different layers -> wire + via + wire).

## Test
Added a unit test: a single-point `hdRoute` with two distinct same-layer `connectionPoints` now yields a 2-point connecting route. Fails on `main` (single-point output), passes with the fix.

Fixes #1528.

### DIFF
--- a/lib/utils/convertHdRouteToSimplifiedRoute.ts
+++ b/lib/utils/convertHdRouteToSimplifiedRoute.ts
@@ -343,6 +343,54 @@ export const convertHdRouteToSimplifiedRoute = (
     }
   }
 
+  // Same-region connection fallback. When both endpoints of a connection fall in
+  // a single capacity node, no region boundary is crossed. The terminal port
+  // points are filtered out of the HD node graph (buildInputNodesWithPortPoints
+  // drops `_tinyTerminal` ports because they are normally attached only at route
+  // ends), so the HD solver returns a degenerate single-point route and the net
+  // is left unconnected - e.g. two adjacent same-net pads in one node. Emit the
+  // direct intra-node segment between the two endpoints so the connection is
+  // actually made.
+  const distinctWireXy = new Set(
+    result
+      .filter((seg) => seg.route_type === "wire")
+      .map((seg) => `${(seg as { x: number }).x},${(seg as { y: number }).y}`),
+  )
+  const cps = opts.connectionPoints
+  if (cps && cps.length >= 2 && distinctWireXy.size < 2) {
+    const a = cps[0]
+    const b = cps[cps.length - 1]
+    if (
+      isSingleLayerConnectionPoint(a) &&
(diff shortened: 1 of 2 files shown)


maci0:
Reopen request: auto-closed by the 1-day stale bot. Fixes degenerate same-region HD routes (direct segment for a same-region connection). Related to the same-layer-short work in #1507 / #1513. CI-green; branch intact.

maci0:
Reopened against current main as a fresh PR (this one was stale-bot closed).

maci0:
New PR: https://github.com/tscircuit/tscircuit-autorouter/pull/2561

PR #1531, opened 2026-07-07
fix: guard getEffortScale against undefined/NaN effort (NaN iteration cap)

### LINKED ISSUE #1525: getEffortScale returns NaN for undefined effort -> MAX_ITERATIONS NaN -> instant 'ran out of iterations'
## Root cause
```ts
const getEffortScale = (effort: number) => Math.max(effort, 1e-2)
```
`Math.max(undefined, 1e-2)` is **NaN**. When `effort` is undefined, every `MAX_ITERATIONS: Math.ceil(N * getEffortScale(effort))` becomes `NaN`, and the solver reports `TinyHyperGraphSolver ran out of iterations` on the **first** step (the `iterations >= NaN` guard).

## When effort is undefined
`TinyHypergraphPortPointPathingSolver` runs the `DuplicateCongestedPortSolver` prepass when `params.connections.length <= MAX_CONNECTIONS_FOR_DUPLICATE_CONGESTED_PORT_PREPASS`, and passes `getEffortScale(params.effort)`:

```ts
MAX_ITERATIONS: Math.ceil(10_000_000 * getEffortScale(params.effort)),
```

`AutoroutingPipelineSolver4` defaults its own `this.effort = mutableOpts.effort ?? 1`, but that default is **not threaded into** `HgPortPointPathingSolverParams.effort`, so `params.effort` here is `undefined`. Boards with few connections (which hit the prepass) therefore NaN out; larger boards that skip the prepass are unaffected (e.g. the `dataset-srj15 sample11` fixture passes, masking this).

## Symptom
A small/low-connection 2-layer board that should route in seconds instead fails immediately with `ran out of iterations`, 0 traces produced.

## Repro
```js
Math.max(undefined, 1e-2)   // => NaN
```
Route any board whose `connections.length` is under the prepass threshold with no explicit `effort`; it instant-fails. Forcing `getEffortScale` to floor undefined to a real number (`Math.max(Number(effort) || 1, 1e-2)`) makes the same board route.

## Fix options
- Make `getEffortScale` undefined-safe: `Math.max(Number(effort) || 1, 1e-2)`, or
- Thread `AutoroutingPipelineSolver4.effort` into `HgPortPointPathingSolverParams.effort` (and the other params objects) so it is never undefined.

maci0:
## Problem
`getEffortScale = (effort) => Math.max(effort, 1e-2)` returns **NaN** when `effort` is undefined (`Math.max(undefined, 1e-2) === NaN`). `effort` is optional and arrives undefined on boards that hit the `DuplicateCongestedPortSolver` prepass (`params.effort` is not threaded from the pipeline's `effort ?? 1` default). Every `MAX_ITERATIONS = N * getEffortScale(effort)` then becomes NaN, and the solver fails on the first step with `TinyHyperGraphSolver ran out of iterations`, producing 0 traces. Larger boards that skip the prepass are unaffected, which masks it. (#1525)

## Fix
Floor undefined/NaN to a real default: `Math.max(Number(effort) || 1, 1e-2)`. Exported for a regression test.

## Test
`getEffortScale(undefined)` is finite and positive (and the derived iteration cap is finite); explicit efforts are unchanged.

Fixes #1525.

### DIFF
--- a/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
+++ b/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
@@ -113,7 +113,8 @@ const DUPLICATE_PORT_TRAVERSAL_PENALTY = 150
 const CRAMPED_PORT_TRAVERSAL_PENALTY = 150
 const MAX_CONNECTIONS_FOR_DUPLICATE_CONGESTED_PORT_PREPASS = 180
 
-export const getEffortScale = (effort: number) => Math.max(Number(effort) || 1, 1e-2)
+export const getEffortScale = (effort: number) =>
+  Math.max(Number(effort) || 1, 1e-2)
 
 const getTinyViaSizeOptions = (
   minViaPadDiameter?: number,
--- a/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
+++ b/lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver.ts
@@ -113,7 +113,7 @@ const DUPLICATE_PORT_TRAVERSAL_PENALTY = 150
 const CRAMPED_PORT_TRAVERSAL_PENALTY = 150
 const MAX_CONNECTIONS_FOR_DUPLICATE_CONGESTED_PORT_PREPASS = 180
 
-const getEffortScale = (effort: number) => Math.max(effort, 1e-2)
+export const getEffortScale = (effort: number) => Math.max(Number(effort) || 1, 1e-2)
 
 const getTinyViaSizeOptions = (
   minViaPadDiameter?: number,
--- a/tests/solvers/getEffortScale-undefined.test.ts
+++ b/tests/solvers/getEffortScale-undefined.test.ts
@@ -1,0 +1,20 @@
+import { expect, test } from "bun:test"
+import { getEffortScale } from "lib/solvers/PortPointPathingSolver/tinyhypergraph/TinyHypergraphPortPointPathingSolver"
+
+// Regression: effort is optional and reaches getEffortScale as `undefined` on
+// low-connection boards (the DuplicateCongestedPortSolver prepass passes
+// params.effort straight through). `Math.max(undefined, 1e-2)` is NaN, which
+// makes every `MAX_ITERATIONS = N * getEffortScale(effort)` NaN and the solver
+// reports "ran out of iterations" on the first step. getEffortScale must return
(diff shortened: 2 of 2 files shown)


0hmX:
Do you have the board can be tsx or an auto router bug report not sure if it's possible to hit or srj

maci0:
@0hmX good instinct to push on this. I dug in on the published `@tscircuit/capacity-autorouter@0.0.654` and I **cannot** hit the NaN path from a SimpleRouteJson through the public pipeline solvers. Your doubt was right, so let me lay out what I actually found rather than hand-wave.

The footgun in `getEffortScale` is real at the unit level:

```js
Math.max(undefined, 1e-2) // => NaN
```

But it is not reachable from an SRJ, because both pipelines that use the TinyHypergraph solver default and thread `effort` before it ever reaches `getEffortScale`. In the 0.0.654 bundle:

- pipeline ctor: `effort = opts.effort ?? 1`
- tiny-solver params: `{ ..., effort: cms.effort, ... }`

So `params.effort` is `1` from any SRJ entry, never `undefined`. A low-connection board (3 connections, well under the 180 prepass threshold, so it does run the `DuplicateCongestedPortSolver` prepass) routes fine with no effort.

I also forced the failure the issue describes by passing `effort: NaN` explicitly (`?? 1` only catches null/undefined, not NaN). That does make `MAX_ITERATIONS` NaN, but the board **still routes** and never reports `ran out of iterations`, because the guards are:

```js
!this.solved && this.iterations >= this.MAX_ITERATIONS   // >= NaN  -> false
!this.solved && this.iterations >  this.MAX_ITERATIONS   // >  NaN  -> false
```

`iterations >= NaN` / `iterations > NaN` are both `false`, so a NaN cap doesn't trip the exhaustion path at all, it just removes the ceiling.

Run against 0.0.654 ([gist](https://gist.github.com/maci0/48f890f76bea23438a4980e9b808a402)):

```
Math.max(undefined, 1e-2) = NaN
GY  no opts            solved=true failed=false traces=3 MAX_ITERATIONS=100000000
GY  effort: undefined  solved=true failed=false traces=3 MAX_ITERATIONS=100000000
GY  effort: 1          solved=true failed=false traces=3 MAX_ITERATIONS=100000000
GY  effort: NaN        solved=true failed=false traces=3 MAX_ITERATIONS=NaN
CW  ... (Pipeline7 identical) ...
```

So: on 0.0.654 this is a latent footgun, not a reachable-from-SRJ failure. The only ways to a NaN cap are constructing `TinyHypergraphPortPointPathingSolver` directly with `effort` omitted (not exported, not an SRJ path) or passing `effort: NaN` explicitly. I don't have an SRJ that fails, so I'd treat this PR as cheap defense-in-depth on `getEffortScale` rather than a fix for an observable routing bug, and it's totally fine to close it if you'd rather not carry the guard. Sorry for the noise on the original issue framing.

PR #1536, opened 2026-07-07
Fix stitch endpoint snapping to prevent closed-loop two-port traces

AnasSarkiz:
## Summary

Fixes a route-stitching bug where a normal two-port connection could accidentally be stitched from one port back to the same port.

This caused closed-loop traces like:

port A -> route geometry -> port B -> port A

Later stages treated the trace as starting and ending at `port A`, so DRC reported `port B` as missing a connection.

## Root Cause

`snapIslandEndpointToNearestTerminal` sorted the terminal list to choose the initial closest terminal, but then continued checking `params.terminals.slice(1)` from the original unsorted list.

In `dataset01` circuit107, this skipped the real second terminal and snapped both route endpoints to `pcb_port_27`.

## Fix

Use the same sorted terminal list for the full nearest-terminal search.

This keeps the two endpoints separate, so the stitcher outputs an open route between the real ports instead of a closed loop.

Screenshot 2026-07-07 at 5 46 46 PM
Screenshot 2026-07-07 at 5 44 11 PM
Screenshot 2026-07-07 at 5 42 43 PM


seveibar (approved):

0hmX (approved):

AnasSarkiz:
Screenshot 2026-07-07 at 9 07 46 PM

seveibar:
@copilot fix the failing snapshots

seveibar:
@copilot fix the tests/features/pipeline4-dataset01-circuit015-cmn2-high-density-only.test.ts test

PR #1545, opened 2026-07-08
Fix BGA grid alignment for obstacle target routing

ShiboSoftwareDev:
Preserves observed BGA pad centers and uses adjacent spacing for gap nodes so mesh
  regions border cleanly. Marks obstacle-backed targets with connection metadata and
  requires normal bordering routing edges instead of nearest-node fallback. Extracts
  tiny-route net indexing helpers and serializes region net ids to keep target
  region ownership explicit.
image

image


ShiboSoftwareDev:
/benchmark --pipeline 7 --dataset 18 --sample-timeout 2000s

lib/solvers/PortPointPathingSolver/tinyhypergraph/createTinyRouteNetIndexer.ts
```diff
@@ -0,0 +1,23 @@
+export type TinyRouteNetIndexSource = {
+  connectionId: string
+  mutuallyConnectedNetworkId?: string
+}
+
+export type TinyRouteNetIndexer = (
+  routeMetadata: TinyRouteNetIndexSource,
+) => number
+
+export function createTinyRouteNetIndexer(): TinyRouteNetIndexer {
```
0hmX, line 10: I think you have two PRs here instead of one. This is mixing BGA escape changes with Net ID changes. Breaking it up into two PRs will also give us better visibility into what is helping the most.

lib/types/capacity-mesh-types.ts
```diff
@@ -40,5 +40,6 @@ export interface CapacityMeshEdge {
   capacityMeshEdgeId: string
   nodeIds: [CapacityMeshNodeId, CapacityMeshNodeId]
   isOffboardEdge?: boolean
+  isTargetEscapeEdge?: boolean
```
0hmX: Add some JSDoc here. This fix is hiding the bug where we have gaps between regions. I’m happy with merging this, but we need to write this down so we can maybe fix it later. Please add a small JSDoc comment explaining this.

lib/solvers/PortPointPathingSolver/tinyhypergraph/getRegionNetIdByRegionId.ts
```diff
@@ -0,0 +1,45 @@
+import type { HgPortPointPathingSolverParams } from "../hgportpointpathingsolver/types"
+import { checkIfConnectionPointIsInRegion } from "../hgportpointpathingsolver/checkIfConnectionPointIsInRegion"
+import type { TinyRouteNetIndexer } from "./createTinyRouteNetIndexer"
+
+export function getRegionNetIdByRegionId(input: {
```
0hmX, line 5: i think this should be a separate pr

0hmX (changes requested):

lib/solvers/PortPointPathingSolver/tinyhypergraph/getRegionNetIdByRegionId.ts
ShiboSoftwareDev: @0hmX  we need both changes to make it work don't you think, I say we should address the gap issue and maybe come up with a better method, but let's keep it to one PR, wdyt?

ShiboSoftwareDev:
/benchmark --pipeline 7 --dataset 18 --sample-timeout 2000s

0hmX (approved):

seveibar (approved):

PR #1552, opened 2026-07-08
Fix same-net via merger route transitions

0hmX:
## Summary
- allow same-net route geometry to be ignored during near via merge collision checks
- move the existing via transition route-point cluster to the kept via instead of inserting a second transition
- update bugreport71 snapshots for the corrected trace simplification output

## Tests
- bun test tests/solvers/same-net-via-merger-near-same-net-route-conflict.test.ts tests/solvers/same-net-via-merger-direct-overlap-star.test.ts tests/bugs/bugreport71-dd7d15.test.ts --timeout 9999999


0hmX:
@copilot bugreport68-solar-battery-charger.srj.json update please this one only update bugreport68-solar-battery-charger.srj.json this test snaphtos noting else; use --timeout 999999999

seveibar (approved):

PR #1553, opened 2026-07-08
Make pipeline 7 GlobalDrc connMap aware

0hmX:
to: https://github.com/tscircuit/high-density-repair03/pull/23

## Summary
- keep pipeline 7 using the existing GlobalDrcForceImproveSolver class
- normalize the GlobalDrc constructor SRJ and HD routes with cms.connMap so same-net vias, pads, and route geometry share net identity inside GlobalDrc
- restore original route root names when reading GlobalDrc output from pipeline 7

## Tests
- bun test tests/solvers/pipeline7-global-drc-conn-map-inputs.test.ts tests/bugs/bugreport71-dd7d15.test.ts --timeout 9999999
- bun run build


AnasSarkiz (approved):
Great results

seveibar (approved):

PR #1572, opened 2026-07-09
ci: skip version bump PR checks

0hmX:
## Summary
- skip regular PR CI jobs when the PR head branch starts with `version-bumps/`
- skip the benchmark instructions comment for automated version bump PRs
- add a Vercel ignore command so external preview deployments are ignored for `version-bumps/**` branches

## Logic / fallacy review
- The skip predicate is branch-scoped, not path-scoped. This avoids the false equivalence that every `package.json` change is an automated version bump.
- The GitHub Actions guards keep `push` and manual/non-PR behavior unchanged by checking `github.event_name` before evaluating PR branch context.
- The `pull_request_target` workflow uses `github.event.pull_request.head.ref` because `github.head_ref` is not the reliable context for that event.
- The Vercel rule follows Vercel's documented ignore-command semantics: exit `0` ignores the deployment and exit `1` continues it.
- Residual risk: if a human intentionally uses a `version-bumps/` branch for non-version work, these checks will skip. That matches the repository's existing automated branch convention.

## Validation
- parsed changed workflow YAML with the repo-available `yaml` package
- parsed `vercel.json`
- ran `git diff --check`

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,0 +1,4 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "ignoreCommand": "case \"$VERCEL_GIT_COMMIT_REF\" in version-bumps/*) exit 0 ;; *) exit 1 ;; esac"
+}
--- a/vercel.json
+++ b/vercel.json
@@ -1,4 +1,0 @@
-{
-  "$schema": "https://openapi.vercel.sh/vercel.json",
-  "ignoreCommand": "case \"$VERCEL_GIT_COMMIT_REF\" in version-bumps/*) exit 0 ;; *) exit 1 ;; esac"
-}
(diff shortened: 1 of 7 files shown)


0hmX (commented):
Logic/fallacy review: I do not see a logic fallacy in the skip condition. The rule targets the automated branch convention, not the changed file set, so it avoids overgeneralizing all package metadata edits as safe to skip. Event contexts are separated correctly: pull_request jobs use github.head_ref, while pull_request_target uses github.event.pull_request.head.ref. The Vercel ignore command uses the documented exit-code polarity: 0 to ignore, 1 to continue. Main residual risk is deliberate non-version work on a version-bumps/* branch, which is outside the intended branch convention.

vercel.json
imrishabh18, line 1: what is this for?

imrishabh18 (approved):

seveibar:
@0hmX what’s the motivation?

seveibar (approved):
I dont know if its better to skip or to fix code sync

0hmX:
> I dont know if its better to skip or to fix code sync

i will merge this for now but i am not sure what you mean do you mean to use git tags for version?

PR #1580, opened 2026-07-10
Add 45-degree-trace SRJ23 benchmark dataset 

0hmX:
## Summary
- add the SRJ23 dataset pinned to b4fb684
- add an interactive benchmark fixture for its circuit exports
- enable SRJ23 in local and workflow benchmark dataset selection

## Validation
- `bun test tests/benchmark-datasets.test.ts --timeout 9999999`
- `./benchmark.sh --dataset srj23 --scenario-limit 1 --concurrency 1`


0hmX:
image

PR #1592, opened 2026-07-10
Add a solver to merge global and component topologies in Pipeline 7 & consistently use colors for layers

ShiboSoftwareDev:
- Adds a layer-aware solver that overlays global and component topology groups
    into one routing mesh.

  - Refines only intersections between different groups, preserving intentional
    same-group overlaps and exact global topology when no components exist.

  - Restores authoritative target regions and tracks component provenance,
    preventing TinyHypergraph endpoint disconnections.

  - Makes component obstacles layer-aware and adds dataset01, dataset18, srj21, and
    focused topology regression coverage.
image


seveibar (commented):
Going to do a morenin depth review soon- but this pr removes net coloring- im not completely sure why because net coloring is useful in some contexts- is it permanently removed? Are we going to regret removing it? Layer coloring is good for simplification but i dont know if we want to completely remove net coloring

seveibar:
@ShiboSoftwareDev i dont get the net coloring change tbh, consistency is good but removing the ability to see a net? Shpuldnt be in this pr. Big change and obfuscating the diff!!

seveibar (approved):
Make sure to remove idea.md cc @0hmX for thoughts

0hmX (approved):

PR #1601, opened 2026-07-12
Skip pre-routed trace connections in Pipeline7

seveibar:
https://github.com/user-attachments/assets/00d7fce5-bb55-4fa8-bf64-514f7acbf9e1

## Summary

- infer connected point pairs from pre-routed trace endpoints when an SRJ omits `connectsTo`
- pass pre-connected point groups into the MST builder, which unions them before selecting new edges
- keep explicit `connectsTo` metadata authoritative
- add a Pipeline7 regression test using dataset-srj23 circuit015

## Root cause

Pipeline7 only recognized initial traces as already connected when they explicitly listed at least two point IDs in `connectsTo`. Dataset-srj23 initial traces omit that field, even though their `connection_name` and route endpoints identify the completed connection. The point-pair stage therefore scheduled those connections again.

## Impact

Across all 107 dataset-srj23 samples, point-pair routing jobs decrease from 3,211 to 2,108 (1,103 fewer, or 34.35%). Circuit015 emits one new trace instead of five.

Blacksmith benchmark, Pipeline7, dataset-srj23, 1x effort, 4 workers:

| Metric | Before | After |
| --- | ---: | ---: |
| Completed | 48.6% (52/107) | 58.9% (63/107) |
| Relaxed DRC pass | 15.9% (17/107) | 22.4% (24/107) |
| P50 time | 2.9s | 1.9s |
| Average vias | 24.27 | 20.05 |

## Validation

- `bun test tests/features/preexisting-connected-traces tests/features/pipeline7-dataset-srj23-preexisting-traces.test.ts --timeout 9999999` (7 passed)
- `bun test tests/offboard-substitution --timeout 9999999` (5 passed)
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`
- post-refactor dataset-srj23 point-pair count: 3,211 before, 2,108 after
- full before/after Blacksmith benchmark described above


seveibar:
@0hmX i think this may be a dataset issue- was hoping you could investigate that possibility- according to the ai description it seems to be a dataset issue- so we should avoid making the code more complex

PR #1602, opened 2026-07-13
Emit connectivity metadata from Pipeline7 traces

0hmX:
## Summary
- Emit `connectsTo` point IDs on each generated Pipeline7 PCB trace.
- Preserve the exact `pointId` values, including optimized off-board endpoints.
- Throw loudly if a purported point pair contains more than two points.

This complements #1601: it accepts older traces without metadata, while this PR makes newly generated traces self-describing.

## Validation
- TypeScript check
- Full dataset01/Pipeline7 benchmark with four workers: 85 of 85 completed, no timeouts, 88.2% relaxed-DRC pass, P50 1.8s, P95 9.9s.
- Diff check

The package build still fails before bundling because tsup cannot resolve existing lib path aliases. Declaration generation succeeds; the failure is unrelated to this change.


0hmX:
@copilot resolve the merge conflicts in this pull request

PR #1626, opened 2026-07-14
Preserve undersized topology bridges

ShiboSoftwareDev:
sample 15 now solves:
image

## Summary

- preserve low-area RectDiff nodes when they connect a target or bridge at least two same-layer regions
- continue pruning sub-0.002 mm coordinate artifacts
- add a focused regression test for bridge preservation and numerical-sliver removal

## Root cause

Dataset 18 sample 15 contains a 0.9865 mm by 0.006 mm RectDiff slice directly between `cmn_881` and the multilayer region `cmn_176`. `NodeDimensionSubdivisionSolver` removed it because its area was below 0.01 mm².

Although the slice is narrow, it is a connectivity bridge: removing it makes `source_trace_57__source_net_57_mst2` statically unreachable. Selective rerip cannot repair immutable topology, so it repeatedly rebuilt the other routes until its iteration limit was exhausted.

This PR is stacked on #1625 so it uses the selective-rerip starvation fix pinned there.

## Validation

- `bun test tests/node-dimension-subdivision-solver.test.ts tests/node-dimension-subdivision-connectivity-bridge.test.ts --timeout 9999999` — 8 passed
- `bun run build` — passed
- Pipeline 7, dataset 18, effort 1x, sample 15 — completed in 192.8s with 324 vias; relaxed DRC remains failing as in the baseline
- Pipeline 7, dataset 18, effort 1x, sample 5 — completed in 21.2s with 169 vias; relaxed DRC passed
- Pipeline 7, dataset 18, effort 1x, sample 1 — completed in 22.2s with 195 vias; its two clearance reports exactly match an explicit run of the parent branch


0hmX (approved):
for dataset 18 100% :heart: :heart:

AnasSarkiz (approved):

PR #1631, opened 2026-07-14
Improve benchmark snapshot viewer

seveibar:
## Summary

- show trace counts and DRC issue counts on every benchmark snapshot card
- make the existing fullscreen control fill the browser viewport without entering browser fullscreen mode
- keep zoomed snapshots sharp by zooming the inline SVG through its `viewBox`
- remove connection-point/port markers from snapshot SVGs

## Why

The snapshot page did not expose basic routing complexity or the number of DRC findings. Its fullscreen control also entered browser fullscreen, and CSS-transform zooming could cause the SVG to be rasterized and look blurry. Connection-point markers added visual noise to routed-output snapshots.

## Validation

- `bun test tests/benchmark-snapshots-html.test.ts --timeout 9999999`
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`

### DIFF
--- a/scripts/benchmark/benchmark-run-task.ts
+++ b/scripts/benchmark/benchmark-run-task.ts
@@ -372,6 +372,7 @@ const createBenchmarkSnapshot = async ({
     sampleNumber: task.sampleNumber,
     label: `${task.datasetName} sample ${task.sampleNumber} - ${task.solverName}`,
     elapsedTimeMs,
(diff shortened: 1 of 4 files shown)


ShiboSoftwareDev (approved):

ShiboSoftwareDev:
the background is not zooming correctly with the circuit
image

PR #1632, opened 2026-07-14
Expand high-density routing portfolio dynamically

ShiboSoftwareDev:
image

image

## What changed

- remove Pipeline 7's fixed 8,000-round inner high-density cap
- initialize candidates through their idempotent setup lifecycle so their problem- and effort-derived limits are known without advancing them
- derive the supervisor limit from every live candidate's remaining work
- preserve the existing ordering portfolio as the initial search
- expand into the remaining deterministic ordering variants only after the initial portfolio consumes an aggregate exploration budget equal to its most expensive candidate's natural budget, or exhausts
- leave candidate selection to the existing fitness scheduler
- add a focused regression fixture and tests for the large sample 2 node

## Why

The sample 2 node is routable at its original physical size. The fixed outer cap stopped the portfolio before a viable ordering completed; grow/shrink later returned invalid direct connections, producing long intersecting traces.

An eager six-ordering portfolio removed those malformed traces, but changed winners on easy boards and regressed sample 5 from a DRC pass to a failure. This staged policy keeps the original portfolio behavior first, scales additional search breadth from the candidates' own budgets instead of a fixed iteration count, and never directly selects or forces a solver.

## Validation

Dataset 18, effort 1x:

| Sample | Completion | Relaxed DRC | High-density stage |
| --- | --- | --- | ---: |
| 1 | pass | fail, 2 errors (matches baseline) | 9.2s |
| 2 | pass | fail, 72 errors | 105.5s |
| 5 | pass | pass, 0 errors (matches baseline) | 10.8s |
| 6 | pass | fail, 506 errors | 257.1s |

- sample 2's butchered cross-board traces are visually removed
- the captured hard node solves at its original size with 19 routes, without resize or invalid fallback output
- 23 focused tests pass
- production TypeScript and declaration build passes
- formatting, lint, and diff checks pass

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
@@ -540,7 +540,6 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
           obstacles: cms.srj.obstacles,
           layerCount: cms.srj.layerCount,
           useGrowShrinkHighDensityIntraNodeSolver: true,
-          growShrinkMaxInnerIterationsPerGrowthAttempt: 8_000,
           growShrinkFallbackToInvalidGeometryOnFailure: true,
         },
       ]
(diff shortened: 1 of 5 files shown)


0hmX (approved):

seveibar (approved):
I wouldn't call this a Hyper* solver, just because a Hyper solver is a bit more predictable/formal in how it allocates iterations. We could potentially call this like a PortfolioSingleIntraNodeSolver (or something like that) just to disambiguate, but it's good

PR #1659, opened 2026-07-15
[156 DRC fixed in dataset 18] repair dense-terminal DRCs with via-in-pad transitions

ShiboSoftwareDev:
## Summary

- lock route terminals to their authoritative PCB port coordinates before post-route repair
- run DRC-scored via-in-pad topology repair inside `exactGeometryDrcForceImproveSolver`
- relocate terminal-side vias and move single-layer terminal routes using pad-centered transitions
- pass the board's via drill diameter to the repair solver and score every candidate with the complete DRC result
- pin `high-density-repair03` to tscircuit/high-density-repair03#26 commit `adb38a0`

Depends on tscircuit/high-density-repair03#26.

## Why

Dataset 18's remaining failures are concentrated in dense terminal fields. Many routes leave a terminal across neighboring surface pads and change layers only afterward. Point-force repair can move geometry, but it cannot move that layer transition into the connected terminal pad.

The exact-geometry DRC solver now owns that topology repair as its final internal phase. It can either relocate an existing terminal-side via to the connected pad or move a fully single-layer terminal route to another layer with transitions at both pad centers. `createViaInPadDrcEvaluator` remains the autorouter adapter that evaluates candidates using the complete board DRC result; candidates with a higher DRC count are rejected.

## Benchmark regression fixed

Comparing the supplied Dataset 18 artifacts exposed a regression from 442 to 548 total DRCs. The eligibility guard incorrectly required the 0.30 mm outer via land to fit inside 0.20–0.25 mm terminal pads. These boards use a 0.15 mm drill: the same-net copper land may extend past the pad, while the drilled hole must remain inside the pad copper.

The autorouter now passes `viaHoleDiameter` to the solver, and the solver validates the drill rather than the outer via land.

The authoritative Blacksmith run on commit `f67bc028` completed all 16 samples:

| Result | Total DRCs | Total sample time | Total vias |
|---|---:|---:|---:|
| Earlier supplied artifact | 442 | 6194.9 s | 4153 |
| Regressed supplied artifact | 548 | 7530.6 s | 3617 |
| Fixed Blacksmith run | **421** | **5827.7 s** | 3639 |

That removes 127 DRCs (23.2%) from the regressed run and finishes 21 DRCs below the earlier artifact. It also reduces summed sample time by 1702.9 seconds compared with the regressed artifact.

The affected focused samples in the same artifact comparison:

| Sample | Earlier artifact | Regressed artifact | Fixed locally |
|---:|---:|---:|---:|
| 13 | 56 | 105 | 54 |
| 14 | 93 | 138 | 93 |
| 16 | 6 | 5 | 5 |

The two major regressions fall from 243 to 147 DRCs. Sample 14 fully returns to its earlier result, sample 13 is two DRCs better, and sample 16 remains at 5.

[View the Blacksmith benchmark run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/29517726665).

## Validation

- focused autorouter tests — 3 pass, 0 fail
- `bun run build` — pass, including declaration generation
- `git diff --check` — pass
- dependency suite — 29 pass, 0 fail
- dependency typecheck and format check — pass
- full Dataset 18 Blacksmith benchmark — 421 total DRCs, all 16 samples completed
- Dataset 18 samples 13, 14, and 16 — 54, 93, and 5 DRCs respectively


seveibar (approved):
name?

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/lock-hd-route-terminals.ts
```diff
@@ -0,0 +1,78 @@
...
+): HighDensityRoute[] => {
+  const connectionByName = new Map(
+    connections.map((connection) => [connection.name, connection]),
+  )
+
+  return hdRoutes.map((hdRoute) => {
+    if (hdRoute.route.length === 0) return hdRoute
+
+    const connection = connectionByName.get(hdRoute.connectionName)
+    if (!connection || connection.pointsToConnect.length !== 2) return hdRoute
```
0hmX, line 22: Do you think this should throw; invalid data should throw by default i think

0hmX (approved):
lgtm

AnasSarkiz:
why the SRJ20 results getting this bad??? 
`Completed %` from `71.5% ` to `36.0%`
High P50 and many timeouts

ShiboSoftwareDev:
> why the SRJ20 results getting this bad??? `Completed %` from `71.5% ` to `36.0%` High P50 and many timeouts

true I didn't notice

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/lock-hd-route-terminals.ts
```diff
@@ -0,0 +1,78 @@
...
+    const reversedDistance =
+      Math.hypot(
+        firstRoutePoint.x - secondConnectionPoint!.x,
+        firstRoutePoint.y - secondConnectionPoint!.y,
+      ) +
+      Math.hypot(
+        lastRoutePoint.x - firstConnectionPoint!.x,
+        lastRoutePoint.y - firstConnectionPoint!.y,
+      )
+    const [startTerminal, endTerminal] =
```
0hmX, line 46: I’d love it if this weren’t based on a distance check and could instead be set during the port-point pathing stage.

i think its the first points were you can get this data

seveibar (approved):

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/lock-hd-route-terminals.ts
seveibar: (this seems worth looking into!!!!)

AnasSarkiz (approved):

AnasSarkiz:
> > why the SRJ20 results getting this bad??? `Completed %` from `71.5% ` to `36.0%` High P50 and many timeouts
> 
> true I didn't notice

I think it was blacksmith issue

PR #1675, opened 2026-07-16
Add long-running PR benchmark option

ShiboSoftwareDev:
## Summary

- increase the regular autorouting benchmark timeout from 60 to 120 minutes
- add `/benchmark-long [benchmark.sh args...]` for long-running PR benchmarks
- run long benchmarks on `blacksmith-8vcpu-ubuntu-2404-arm` with 8 workers by default and a 480-minute timeout
- document the new command in PR benchmark instructions

## Why

PR comment benchmarks currently share one runner configuration and a one-hour timeout. Larger datasets such as srj18 can exhaust that window while using only four workers on the configured runner. The new command provides an explicit higher-capacity option without making every benchmark an eight-hour job.

`/benchmark-long` accepts the same arguments as `/benchmark`; for example:

```text
/benchmark-long --dataset srj18
```

Callers can still override `--concurrency` when needed.

## Validation

- `bun test tests/pr-benchmark-command.test.ts --timeout 9999999`
- `bun run format:check`
- parsed the edited workflow files with Ruby's YAML parser
- `git diff --check`


0hmX (approved):

.github/workflows/benchmark.yml
```diff
@@ -67,8 +72,8 @@ jobs:
         github.event_name == 'pull_request' &&
         contains(github.event.pull_request.title, '[BENCHMARK TEST]')
       )
-    runs-on: ${{ vars.BENCHMARK_RUNNER || 'blacksmith-32vcpu-ubuntu-2404-arm' }}
-    timeout-minutes: 60
+    runs-on: ${{ inputs.long_benchmark && 'blacksmith-8vcpu-ubuntu-2404-arm' || vars.BENCHMARK_RUNNER || 'blacksmith-32vcpu-ubuntu-2404-arm' }}
+    timeout-minutes: ${{ inputs.long_benchmark && 240 || 120 }}
```
seveibar:
```suggestion
    timeout-minutes: ${{ inputs.long_benchmark && 480 || 120 }}
```

seveibar (approved):

Issue #129, opened 2025-06-05
Autorouting Bug Report: <board#409 />


imrishabh18:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=171bae09-5baf-4b77-8544-0fce1fa4fee4

(any additional details? Maybe a screenshot?)

![Image](https://github.com/user-attachments/assets/f6854948-2707-49ed-8336-66b77b31ca88)

PR #1677, opened 2026-07-16
Use completion-first physical trace occupancy routing

AnasSarkiz:
## Summary

- pass SRJ minimum trace width and obstacle clearance through every tiny-hypergraph solve stage
- pin tiny-hypergraph PR #129 at `b28456ce679937de65618051de15855427f59413`
- use a completion-first physical trace-occupancy model instead of a tuned density multiplier
- add focused coverage for the Pipeline 7 occupancy dimensions
- merge current upstream `main` into the PR branch

Depends on tscircuit/tiny-hypergraph#129.

## Foundational fix

The previous density term estimated congestion from distinct-net count and the region's shortest dimension, then added it directly to initial A* cost. That made the result sensitive to topology subdivision rather than actual copper geometry and changed selective-rerip ordering, causing SRJ18 sample 6 to exhaust its routing budget.

The new model estimates shared swept copper area per layer: segment length multiplied by physical trace pitch, normalized by region area. It excludes the first/longest trace corridor so an otherwise empty narrow channel is not penalized. Occupancy scoring activates only after a complete baseline solution is captured, preserving initial completion and making the quality phase baseline-restorable.

## Validation

- `bun test tests/features/pipeline7-trace-occupancy-parameters.test.ts tests/features/pipeline7-dataset-srj18-sample006-selective-rerip.test.ts --timeout 9999999` (2 pass; sample 6 focused stage 19.7s)
- `bun run build`
- tiny-hypergraph: full 98-test suite, typecheck, and format check

## SRJ18 benchmark evidence

Representative samples 1, 5, 7, and 11 complete with no timeouts. Samples 1 and 5 remain DRC-clean. Compared with current main:

| Sample | Main DRC errors | PR DRC errors |
| --- | ---: | ---: |
| 1 | 0 | 0 |
| 5 | 0 | 0 |
| 7 | 7 | 6 |
| 11 | 6 | 4 |
| **Total** | **13** | **10** |

Sample 6 completes end-to-end in 293.5s under a 600s cap. The original regression point is specifically covered: its Tinyhypergraph selective-rerip stage completes in 19.7s rather than failing near 120s.

This PR remains draft until tiny-hypergraph PR #129 is merged and the dependency is repinned to upstream.


AnasSarkiz:
Closing this unmerged occupancy experiment. Its srj18 benchmark reduced some raw error counts but did not increase the relaxed-DRC pass rate, and sample 6 remained outside the standard 360s completion target.

PR #1680, opened 2026-07-17
Preserve net-owned routing regions

0hmX:
ref: https://github.com/tscircuit/tiny-hypergraph/pull/131

It was removing obstacle regions that did not contain a connection endpoint.

But due to component topology generation with rectangles, some obstacle regions do not contain an endpoint but are still assigned to a net. Those net-owned regions must be preserved, because they can contain valid routing ports and edges.


lib/solvers/PortPointPathingSolver/hgportpointpathingsolver/types.ts
```diff
@@ -46,6 +46,10 @@ export type ConnectionHg = Omit<Connection, "startRegion" | "endRegion"> & {
   simpleRouteConnection?: SimpleRouteConnection
 }
 
+export type ConnectionHgWithSimpleRouteConnection = ConnectionHg & {
```
0hmX, line 49: Not required for this fix, but it would make the type API clearer.

We have optional-property paralysis: upstream validates that a value must exist, yet downstream still receives it as optional. `simpleRouteConnection` was an example.

Once a pipeline stage establishes an invariant, its output type should represent that invariant as required. This makes invalid states harder to express and prevents repeated defensive checks downstream.

seveibar (approved):

PR #1688, opened 2026-07-18
Preserve fragmented pad reachability

ShiboSoftwareDev:
## Summary

- preserve route-net identity for connected copper represented by adjacent SRJ obstacle rectangles
- apply the metadata after node-dimension subdivision, and only when an endpoint has no direct routing exit
- add focused unit coverage and an SRJ24 sample 6 static-reachability regression
- run every generated Bun test-plan shard so test files cannot be silently skipped when shard assignments change

## Root cause

The J3 shield polygon is rasterized into adjacent obstacle strips. Node subdivision splits those strips into obstacle-only regions, but the derived regions lose the pad's net identity. The tiny-hypergraph loader then removes the anonymous full-obstacle regions, isolating two GND terminals.

The CI workflow also replaced node 1's generated test plan with a single long-running test. Files assigned to the rest of node 1 were never executed, and adding two tests shifted stale snapshots into node 2.

## Validation

- `bun run build`
- 32 scoped Pipeline 7 tests across 24 files
- SRJ24 sample 6 polygon-pad reachability regression
- existing e2e3 routing snapshots remain unchanged
- generated test-plan audit: 315 files, 315 unique planned entries, none missing or duplicated
- all five Bun Test shards pass on GitHub Actions, including the previously skipped 64-file node-1 plan


seveibar (commented):
Youre not reqlly supposed to process nodes outside a solver

seveibar (commented):
I think this is bad- complex algo without a solver, executed outside a stage

seveibar (approved):

PR #1670, opened 2026-07-15
Superseded: scale post-processing with effort

ShiboSoftwareDev:
## Superseded by focused PRs

This aggregate PR has been replaced by a reviewable dependency stack:

1. #1692 — bound same-net via merging by physical progress
2. #1693 — make trace simplification effort-aware and DRC-safe
3. #1694 — add post-repair trace simplification for effort greater than 1
4. #1695 — add bounded residual local rerouting for effort greater than 1
5. #1698 — cap TinyHypergraph routing work above normal effort
6. #1699 — reserve effort above 1 for post-processing and add force convergence
7. #1700 — scale bounded exact-DRC portfolios with effort

Merge in that order. Each PR is based on the preceding branch and documents its isolated change, effect, convergence behavior, and checks.

The complete replacement stack was run serially on Dataset18 at 100x:

| Sample | This aggregate PR | Complete split stack |
| ---: | --- | --- |
| 1 | 0 DRC / 199 vias / 128.9s | 0 DRC / 188 vias / 59.6s |
| 11 | 0 DRC / 187 vias / 273.2s | 0 DRC / 187 vias / 131.5s |

The original branch and discussion remain available here for history.

PR #1694, opened 2026-07-19
Add DRC-safe post-repair trace simplification

ShiboSoftwareDev:
## Summary

This is part 3 of the focused stack replacing #1670 and is based on the effort-aware trace simplification PR.

Stack: #1692 → #1693 → **#1694** → #1695 → #1698 → #1699 → #1700

Pipeline 7 always runs DRC-checked trace simplification after global DRC repair. Effort changes how deeply the solver searches one ordered strategy portfolio; it does not select a different solver, pipeline topology, or candidate family.

## What changed

- Use `TraceSimplificationSolver` directly as an unconditional Pipeline 7 stage at every positive effort.
- Reconstruct route vias from layer transitions at every effort.
- Preserve and return the best full-board DRC checkpoint instead of the last attempted simplification.
- Stop when route complexity converges or the bounded strategy portfolio is exhausted.
- Scale only the ordered strategy depth: effort 1 gets the first 2 strategies, effort 2 gets the first 3, and higher effort reaches progressively deeper into the same capped portfolio.
- Feed the trace-width result to exact DRC repair whenever that prior stage is available, independent of effort.

There are no conditional stages, effort-only algorithms, or zero-work baseline wrappers.

## Effect

- The `bugreport-b5b3b9d8` baseline regression improves from 3 DRC errors to 2 by eliminating its via-clearance error; its two trace errors remain.
- Existing Pipeline 7 route snapshots are updated where baseline simplification now performs useful work.
- A monotonic checkpoint regression ensures higher effort cannot replace the 1x checkpoint with a dirtier route.

## Checks

- Same Pipeline 7 topology and solver class at effort 1 and effort 2.
- Same via reconstruction and candidate families at effort 1 and effort 2.
- Effort 1 performs work; effort 2 performs at least as much work in the unit fixture.
- 20 targeted stage, snapshot, DRC-count, and direct-stage integration tests passed.
- Serial Dataset 18 post-DRC simplification regression passed in the complete stack.
- `bun run build`
- `git diff --check`


seveibar (changes requested):
Conditional pipeline running is a very bad idea- no conditional stages!

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -84,6 +87,19 @@ interface CapacityMeshSolverOptions {
...
+): Parameters<typeof createPipeline7ExactGeometryDrcEvaluator>[0] => ({
+  connections: solver.netToPointPairsSolver?.newConnections ?? [],
+  originalConnections: solver.originalSrj.connections,
+  layerCount: solver.srj.layerCount,
+  obstacles: solver.srj.obstacles,
+  defaultViaHoleDiameter: solver.viaHoleDiameter,
+  connMap: solver.connMap,
+  srjWithPointPairs: solver.srjWithPointPairs!,
+  originalSrj: solver.originalSrj,
+})
```
seveibar, line 101: move to separate file?

PR #1695, opened 2026-07-19
Add bounded residual local rerouting

ShiboSoftwareDev:
## Summary

This is part 4 of the focused stack replacing #1670 and is based on the post-DRC simplification PR.

Stack: #1692 → #1693 → #1694 → **#1695** → #1698 → #1699 → #1700

Pipeline 7 always runs bounded residual local rerouting for DRC errors left after global repair and simplification. Every positive effort uses the same repair families; effort only expands their numeric work and strategy-depth budgets.

## What changed

- Use `ResidualLocalRerouteSolver` directly as an unconditional Pipeline 7 stage.
- Attribute DRC errors to both participating traces and retain the pad/via geometry required for local repairs.
- Always include directed via shifts, trace shifts, terminal snaps, doglegs, obstacle detours, and layer detours.
- Scale candidate attempts from `ceil(96 * effort)` up to 768 and accepted count-reducing moves from `ceil(2 * effort)` up to 16.
- Let higher effort reach deeper into the same ordered dogleg strategy portfolio.
- Fully validate only locally promising candidates, rank checkpoints count-first, avoid unchanged retries, and stop on convergence or a bounded plateau.

There are no conditional stages, effort-only repair families, or zero-work baseline wrappers.

## Effect

Dataset 18 sample 11 in the focused stage regression:

| Metric | Before residual reroute | After residual reroute |
| --- | ---: | ---: |
| Relaxed DRC errors | 5 | 0 |
| Candidate attempts | — | 102 of a capped 768 |
| Accepted repairs | — | 6 (5 count-reducing, 1 refinement) |

The solver reaches zero relaxed DRC errors without exhausting its cap. Baseline effort also performs bounded repair work; higher effort continues farther only when improvements remain.

## Checks

- 18 residual-reroute and fixed-stage tests passed.
- Serial Dataset 18 residual-reroute regression passed in the complete stack.
- Bugreport69’s localized reroute snapshot passes.
- Bugreport64 and bugreport69 snapshots pass with Linux-native Sharp/resvg, matching CI’s platform.
- `bun run build`
- `git diff --check`


lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/createPipeline7ExactGeometryDrcEvaluator.ts
```diff
@@ -18,10 +18,30 @@ type Pipeline7DrcEvaluatorConversionOptions = Omit<
...
+const getPadIdentifierFromDrcMessage = (
+  message: unknown,
+): string | undefined => {
+  if (typeof message !== "string") return undefined
+  const patterns = [
+    /pcb_smtpad "pcb_port\[#(pcb_port_[^\]]+)\]"/,
+    /Pad pcb_port\[#(pcb_port_[^\]]+)\]/,
+    /smtpad\[#(pcb_port_[^\]]+)\]/,
+    /pcb_plated_hole "pcb_plated_hole\[#(pcb_plated_hole_[^\]]+)\]"/,
+    /Pad pcb_plated_hole\[#(pcb_plated_hole_[^\]]+)\]/,
```
seveibar, line 30: not acceptable

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/createPipeline7ExactGeometryDrcEvaluator.ts
```diff
@@ -31,21 +51,75 @@ const createPipeline7DrcEvaluator = (
...
     )
     const { errors, errorsWithCenters } = getDrcErrors(circuitJson, drcOptions)
+    const viaCenterById = new Map(
+      circuitJson.flatMap((element) =>
+        element.type === "pcb_via"
+          ? [[element.pcb_via_id, { x: element.x, y: element.y }] as const]
+          : [],
+      ),
+    )
+    const padCenterById = new Map<string, { x: number; y: number }>()
```
seveibar, line 72: Never have `Map<string, ...>`

```suggestion
    const padCenterById = new Map<PadId, { x: number; y: number }>()
```

pad id is not a concept we have

lib/solvers/ResidualLocalRerouteSolver/residual-local-reroute-solver.ts
```diff
@@ -0,0 +1,2262 @@
...
+import { BaseSolver } from "../BaseSolver"
+
+type Point = { x: number; y: number }
+
+type Bounds = {
+  minX: number
+  minY: number
+  maxX: number
+  maxY: number
+}
```
seveibar, line 16: redefinition of bounds instead of import, plz fix

lib/solvers/ResidualLocalRerouteSolver/residual-local-reroute-solver.ts
```diff
@@ -0,0 +1,2262 @@
...
+
+type RoutingObstacle = {
+  center: Point
+  width: number
+  height: number
+  connectedTo?: readonly string[]
+}
```
seveibar, line 23: why not using existing type? also unclear this is only rects

PR #1702, opened 2026-07-19
Use rectdiff obstacle connectivity in Pipeline 7

0hmX:
## Summary

- update `@tscircuit/rectdiff` to commit `ba0f5e0` from [rectdiff#138](https://github.com/tscircuit/rectdiff/pull/138)
- remove the Pipeline 7 `FragmentedObstacleConnectivitySolver` stage and reconstruction helpers
- clear connectivity metadata from synthetic component replacement regions before topology merging
- add focused SRJ23 sample 3 and boundary-level connectivity regressions
- retain the SRJ24 sample 6 polygon-pad reachability regression
- update affected visual snapshots

## Why

RectDiff now emits one obstacle-backed mesh node per SRJ obstacle and copies its `connectedTo` metadata directly. Pipeline 7 can consume physical obstacle ownership without rediscovering adjacent fragments.

Synthetic component replacement obstacles are different: their connectivity aliases describe a temporary global-solve placeholder spanning multiple component nets. That metadata must be cleared with the replacement obstacle/target flags before component-local topology is overlaid. Otherwise topology merging unions every component alias into neighboring pad regions, leaving them ambiguously owned and statically unreachable.

## Impact

This removes the Pipeline 7 metadata repair pass while preserving physical obstacle ownership at the SRJ-to-mesh boundary and preventing synthetic replacement metadata from polluting component-local regions.

## Validation

- SRJ23 sample 3 static reachability regression
- SRJ24 sample 6 polygon-pad reachability regression
- focused synthetic replacement metadata unit regression
- affected snapshot tests
- build and type generation


seveibar (approved):
Cc @ShiboSoftwareDev - @0hmX fixed the slop solver that was introduced. Do not use geometric matching to fix data modeling issues!!!

PR #1708, opened 2026-07-20
Merge overlapping external connection groups in autorouting

Abse2001:


lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
```diff
@@ -25,6 +25,7 @@ export const getExternalConnectionState = (
   const allExternalGroups = [...externalGroups, ...routedTraceGroups]
   const pointIdToGroup = new Map<string, number>()
   const pointById = new Map<string, ConnectionPoint>()
+  const parentByPointId = new Map<string, string>()
```
seveibar:
```suggestion
  const parentByPointId = new Map<SomeTypeId, string>()
```

seveibar (changes requested):
wrong place for fix

seveibar:
connMap

seveibar:
Cc @0hmX congused why this is needed- does this indicate a core bug?

PR #1709, opened 2026-07-21
Run low-cost via cleanup after exact DRC repair

ShiboSoftwareDev:
## Summary

- keep Pipeline 7's existing effort-1 routing and DRC repair behavior unchanged
- always run the existing `UselessViaRemovalSolver` and `SameNetViaMergerSolver` once after exact DRC repair
- expose both stages through the normal pipeline visualization/output path
- add a pipeline-order regression test and update the observed DRC-count regression

There are no conditional stages, effort thresholds, new solvers, dependency changes, or duplicated DRC scans in this PR.

## Why this shape

I investigated moving existing simplification work after DRC, rerunning the full simplification loop, and guarding candidates with another exact-geometry DRC scan.

- moving the existing pass improved Dataset 18 quality but made a broad bug-report fixture much slower because DRC received less-simplified routes
- the extra exact DRC checkpoint added about 190 seconds to Dataset 18 sample 11
- the path-simplification phase cost about 6.3 seconds on sample 8 for four additional issues
- via cleanup plus same-net via merging took under 0.9 seconds in the measured sample-1 pipeline and retained the useful low-cost improvement

The final diff therefore leaves all existing work in place and adds only the two cheap existing solver stages after exact repair.

## Serial Dataset 18 benchmark (effort 1)

The PR run used `--concurrency 1 --sample-timeout 10000s` and completed all 16 samples. It produced 403 total DRC issues, 3,508 vias, and 6,655.7 seconds of summed sample time (P50 212.4 seconds, P95 1,116.3 seconds).

The cached main run used a 360-second timeout, so only its nine completed samples can be compared honestly. The other seven PR samples took longer than 360 seconds and would also have timed out under the main run's limit.

| Sample | main DRC | PR DRC | main vias | PR vias | main time | PR time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0 | 0 | 201 | 199 | 82.3s | 81.5s |
| 3 | 30 | 30 | 81 | 80 | 65.9s | 66.5s |
| 4 | 12 | 12 | 152 | 150 | 195.0s | 210.4s |
| 5 | 0 | 0 | 159 | 159 | 71.8s | 71.9s |
| 7 | 7 | 7 | 233 | 232 | 214.8s | 214.3s |
| 9 | 16 | 16 | 114 | 112 | 137.2s | 137.0s |
| 10 | 3 | 2 | 174 | 174 | 188.5s | 202.5s |
| 11 | 6 | 2 | 190 | 190 | 180.5s | 187.3s |
| 16 | 9 | 9 | 110 | 108 | 115.3s | 124.3s |
| **Comparable total** | **83** | **78 (-6.0%)** | **1,414** | **1,404 (-10)** | **1,251.2s** | **1,295.8s (+3.6%)** |

The measured cross-run time difference is small relative to the wall-clock variance between identical runs. The stronger no-overwork signal is the added-stage timing: under 0.9 seconds on sample 1, with no extra DRC scan or effort-scaled loop.

## Validation

- `bun run build`
- focused pipeline-order, final-output, DRC-count, and existing via-merger tests
- full suite: 409 passed, 53 skipped; the remaining failures reproduce on main or are unrelated Pipeline 4/8 long-running snapshot failures
- full serial Dataset 18 effort-1 benchmark: 16/16 completed


ShiboSoftwareDev:
/benchmark-long --dataset srj18 --effort 1 --concurrency 1 --sample-timeout 10000s

PR #1728, opened 2026-07-22
Fix physical net in hypergraph pathing

0hmX:
It’s basically two PRs:

* Merge this PR into the current branch: https://github.com/tscircuit/tscircuit-autorouter/pull/1733
* Put the rest of the changes in this PR.

## Summary

- require connMap-backed canonical physical-net IDs when building hypergraphs
- carry canonical IDs through tiny-hypergraph/BFS routing and fail loudly when they are missing
- canonicalize and deduplicate region connectivity aliases without mutating inputs
- add SRJ23 sample 003 and focused buildHyperGraph regression coverage
- refresh only routing snapshots changed by the corrected net identity

## Root cause

Aliases belonging to the same physical net could receive different routing identities because hypergraph construction used connection/root names instead of the connectivity map. Selective reripping then treated same-net copper as a blocker and repeatedly failed to route SRJ23 sample 003.

## Validation

Before publication:
- production build and declaration generation passed
- TypeScript checking passed
- full test run completed with only expected snapshot changes
- all affected snapshot tests passed after targeted updates
- two independent code reviews found no correctness or scope issues


0hmX:
image


i think most are false postive

seveibar (commented):
remove the concepts of "canonical" and "physical" nets, we only have the following:

- rootConnectionName
- connection name
- connectivity map key (connMapNetworkId?)

AFAIK, there is no need for any other concept. More concepts = exponentially worse complexity

seveibar:
also "canonical", if i wanted to sabotage a codebase, i would keep introducing new terms for the same things, or introduce layers of abstraction where they're not needed

seveibar (approved):

Issue #1721, opened 2026-07-21
Per-connection trace widths are ignored: 1.0 mm motor traces emit as 0.15 mm


seveibar:
## Bug

The autorouter silently emits default-width PCB traces even when each `SimpleRouteConnection` carries a larger `width` and `nominalTraceWidth`. This is visible on an RP2040 + DRV8833 motor-controller board, where the high-current outputs become 0.15 mm traces.

Public reproduction data and PCB snapshot: https://gist.github.com/seveibar/67a25d17202a343a1c8f3d4c47cd06d2

![PCB snapshot](https://gist.githubusercontent.com/seveibar/67a25d17202a343a1c8f3d4c47cd06d2/raw/d5e339087352074601a07d1510f04b8788ec17d4/index.circuit-pcb.snap.svg)

## Exact mismatch

The four motor-output source traces specify `min_trace_thickness: 1`. The generated `SimpleRouteJson` correctly preserves this on connections `source_trace_122` through `source_trace_125`:

```json
{
  "name": "source_trace_122",
  "source_trace_id": "source_trace_122",
  "nominalTraceWidth": 1,
  "width": 1
}
```

All four corresponding output `pcb_trace` routes contain only:

```json
"wire_widths": [0.15]
```

The same issue occurs selectively on the power nets:

- VM terminal branch requests 0.8 mm but outputs 0.15 mm.
- Some merged ground branches request 0.8 mm but output 0.10-0.15 mm.
- Other nearby VM/ground branches do retain 0.8 mm, so the loss appears branch/merge dependent rather than a global parsing failure.

The public gist JSON contains the complete unrouted `simple_route_json`, source constraints, and observed PCB widths.

## Expected behavior

Every wire segment for a connection with `width: 1` / `nominalTraceWidth: 1` should remain 1.0 mm wide. A requested minimum must not silently fall back to the global 0.15 mm width. If the requested width cannot be routed, the route should fail or return an explicit width-degradation error instead.

## Steps to reproduce

1. Download `rp2040-trace-width-bug.json` from the linked public gist.
2. Pass its `simple_route_json` field to the default local autorouting pipeline.
3. Inspect the output routes for `source_trace_122`, `source_trace_123`, `source_trace_124`, and `source_trace_125`.
4. Observe that each wire segment is 0.15 mm instead of the requested 1.0 mm.

## Versions

- `tscircuit@0.0.2116`
- `@tscircuit/core@0.0.1470`
[...]

PR #1736, opened 2026-07-22
fix: preserve per-connection nominalTraceWidth through mergeConnections (#1721 D2)

Hero988:
Partially addresses #1721 (root-cause analysis: https://github.com/tscircuit/tscircuit-autorouter/issues/1721#issuecomment-5038119381)

## Scope

This PR fixes only the **merge-plumbing defect (D2 in the linked analysis)**: per-connection `nominalTraceWidth` was lost whenever connections were merged into a net. It does NOT touch the `TraceWidthSolver` all-or-nothing width schedule that silently falls back to `minTraceWidth` on a single mid-route pinch (D1) — that is a design question awaiting maintainer input, so e.g. the 1.0mm motor traces in the issue's repro that collapse to 0.1mm through a pinched region still do (they route through padstacks that force the narrow schedule). It also leaves the output attribution of merged-net traces to `rootConnectionNames[0]` (D3) unchanged.

## What was broken (D2)

Three stacked losses, all merge-order dependent:

1. `mergeConnections` kept only the **first-found** `nominalTraceWidth` of a merged group ("Take the nominalTraceWidth from the first connection for now").
2. The MST pair connections created by `NetToPointPairsSolver._step` / `NetToPointPairsSolver2_OffBoardConnection._step` **dropped `nominalTraceWidth` entirely**.
3. Pipeline 7 passed the **pre-merge** `srj.connections` to `TraceWidthSolver`, so every `*_mstN` route missed the by-name lookup and resolved via the `rootConnectionName` fallback — i.e. the whole merged net inherited `rootConnectionNames[0]`'s width.

Net effect: in the issue's rp2040 repro, merged nets mixing 0.25mm and 0.8mm (or 0.1 and 0.25) connections all routed at whichever width belonged to the first root.

## Fix

- `mergeConnections` now records each original connection's width + point keys as a `__nominalTraceWidthConstraints` entry on the merged connection (and the merged connection's own `nominalTraceWidth` becomes the group max instead of first-found, so it is order-independent).
- New `getNominalTraceWidthsForMstEdges` resolves a per-edge width over the MST: **every tree edge on the path between a constraint's points must be at least that constraint's width** (max over crossing constraints). Per-endpoint width rules are not sufficient here — the MST can reroute a wide connection's current *through* a narrower connection's endpoint (this exact case occurs in the repro: a 4-connection 0.8/0.25 star where the 0.8 path runs through the 0.25 pad; a min-of-endpoints rule under-widths the through-path edges).
- Both `NetToPointPairsSolver` variants stamp the resolved per-edge width on the pair connections they emit.
- Pipeline 7's `traceWidthSolver` step now passes `[...srj.connections, ...srjWithPointPairs.connections]` so routes resolve widths by their own pair-connection name first; the original connections are kept so the existing `rootConnectionName` fallback behavior is preserved for anything without a per-edge width.

Other pipelines (3/4/6/8, Assignable2) still pass only `srj.connections`; they now benefit from the corrected merged-connection widths but were left unrewired to keep this diff to the default pipeline the issue reproduces on — happy to follow up if you want the same one-liner applied there.

## Tests (red → green)

New, all fail on `main` (pair connections have `nominalTraceWidth: undefined`) and pass with the fix:

- `tests/solvers/net-to-point-pairs-trace-widths.test.ts` — 0.8mm and 0.25mm connections sharing a point: each MST pair connection keeps its own width.
- `tests/solvers/net-to-point-pairs-offboard-trace-widths.test.ts` — same for `NetToPointPairsSolver2_OffBoardConnection` (the solver pipeline 7 actually uses).
- `tests/solvers/net-to-point-pairs-star-trace-widths.test.ts` — the through-path case: MST routes a 0.8mm connection through a 0.25mm connection's endpoint; both edges must be 0.8.

Existing related tests pass: `tests/solvers/merge-connections-missing-roots.test.ts`, `tests/solvers/net-to-point-pairs-root.test.ts`, `tests/features/tracewidthsolver/tracewidthsolver01/02.test.ts`, `tests/features/pour-via-escape/pour-via-escape01*.test.ts` (exercises the other `mergeConnections` consumer). `tsc --noEmit` clean. (I ran targeted test files only — my machine can't run the full suite.)

## Repro evidence (rp2040 board from the issue's gist, default pipeline)

Isolating the `TraceWidthSolver` stage with identical input routes, old wiring vs fixed wiring: **24 routes change width, all increases toward their requested width, zero decreases.** Examples:

- merged net `source_trace_47__source_trace_132__source_net_3` (0.1 + 0.25 + 0.25): before, all 25 MST edges routed at **0.1** — the 0.25 connections' widths were silently discarded; after, the edges on the 0.25 connections' paths get **0.25** (a few 0.175/0.1 where D1's pinch fallback still bites).
- `source_trace_29__87__89` and `source_trace_16__91` (0.1 + 0.15 nets): before all **0.1**, after **0.15** on the 0.15 connections' paths.
- the 0.8/0.25 star `source_trace_109..112`: per-edge requests are now correct (0.8 on all through-path edges) instead of whole-net inheritance from `rootConnectionNames[0]`.

Full-pipeline before/after diff on the repro shows exactly those three merged nets changing and nothing else; the `requested=1 output=[0.1]` motor-trace collapses remain, as expected, because they are D1.

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
@@ -632,7 +632,13 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
         connMap: cms.connMap,
         colorMap: cms.colorMap,
         minTraceWidth: cms.minTraceWidth,
-        connection: cms.srj.connections,
+        // Point-pair connections carry per-edge nominalTraceWidths for merged
+        // nets, so routes resolve widths by their own connection name first.
+        // Original connections are kept for the rootConnectionName fallback.
+        connection: [
+          ...cms.srj.connections,
+          ...cms.srjWithPointPairs!.connections,
+        ] as SimpleRouteConnection[],
         obstacleMargin: cms.srj.minTraceToPadEdgeClearance ?? 0.15,
         layerCount: cms.srj.layerCount,
       },
--- a/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
+++ b/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
@@ -6,7 +6,10 @@ import {
 import { BaseSolver } from "../BaseSolver"
 import { buildMinimumSpanningTree } from "./buildMinimumSpanningTree"
 import { GraphicsObject } from "graphics-debug"
-import { mergeConnections } from "./mergeConnections"
+import {
+  getNominalTraceWidthsForMstEdges,
+  mergeConnections,
+} from "./mergeConnections"
 import { seededRandom } from "lib/utils/cloneAndShuffleArray"
 
 export const getExternalConnectionState = (
@@ -175,9 +178,14 @@ export class NetToPointPairsSolver extends BaseSolver {
     const edges = buildMinimumSpanningTree(connection.pointsToConnect, {
       extraEdges: zeroWeightEdges,
     })
+    const edgeNominalTraceWidths = getNominalTraceWidthsForMstEdges(
+      connection,
+      edges,
+    )
 
     let mstIdx = 0
-    for (const edge of edges) {
+    for (let edgeIndex = 0; edgeIndex < edges.length; edgeIndex++) {
+      const edge = edges[edgeIndex]
       if (areExternallyConnected(pointIdToGroup, edge.from, edge.to)) continue
       this.newConnections.push({
         pointsToConnect: [edge.from, edge.to],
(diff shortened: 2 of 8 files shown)


seveibar (commented):
Can you start with a pr that contains a reproduction using a visual snapshot etc? Im not sure the bug exists

Hero988:
Sure — I'll open a repro-first PR against main: a small board with per-connection trace widths (0.8mm connection sharing a net with a 0.25mm-pad branch) plus a visual snapshot of the routed output, so the width loss is visible directly rather than asserted. For a quick pointer meanwhile: `mergeConnections.ts` L128-135 keeps only the first-found `nominalTraceWidth` for a merged group, and the MST pair connections it generates carry no width at all — so whether a 0.8mm width survives depends on merge order. I'll link the repro PR here shortly and rebase this one on top of it.

Hero988:
Repro PR is up: #1741 — minimal 3-pad board, two connections sharing a pad (0.8mm power run + 0.25mm sense branch on the same net), with an SVG snapshot. On main the routed output contains only 0.25mm wires; the requested 0.8mm never appears, and swapping the order of the two `connections` entries flips which width wins. Once you've had a look I'll rebase this PR on top of it.

seveibar:
it's unclear if this is fully automated. Please make sure you have a repro from an ACTUAL board, synthetic reproductions are not accepted

PR #1738, opened 2026-07-22
fix: handle undefined effort in getEffortScale (#1525)

### LINKED ISSUE #1525: getEffortScale returns NaN for undefined effort -> MAX_ITERATIONS NaN -> instant 'ran out of iterations'

thairc-dev:
## Summary

Closes #1525

When `params.effort` is `undefined`, `Math.max(undefined, 1e-2)` in `getEffortScale` evaluates to `NaN`. This causes `MAX_ITERATIONS` to evaluate to `NaN` across downstream solvers, triggering an immediate 'ran out of iterations' failure on small boards.

This PR updates `getEffortScale` to safely handle `undefined` or non-numeric effort inputs by defaulting to `1`.

## Testing

- Added unit test `tests/get-effort-scale.test.ts` verifying `undefined`, `NaN`, normal, and minimum effort scale calculations.
- All unit tests pass 100%.

### DIFF
--- a/lib/solvers/PortPointPathingSolver/hgportpointpathingsolver/HgPortPointPathingSolverClass.ts
+++ b/lib/solvers/PortPointPathingSolver/hgportpointpathingsolver/HgPortPointPathingSolverClass.ts
@@ -56,7 +56,7 @@ export class HgPortPointPathingSolver extends HyperGraphSolver<
     this.totalRipCount = 0
     if (params.weights.MAX_ITERATIONS_PER_PATH > 0) {
       this.MAX_ITERATIONS =
-        params.weights.MAX_ITERATIONS_PER_PATH * params.effort
+        params.weights.MAX_ITERATIONS_PER_PATH * (params.effort ?? 1)
     }
   }
 
--- a/lib/solvers/PortPointPathingSolver/hgportpointpathingsolver/types.ts
+++ b/lib/solvers/PortPointPathingSolver/hgportpointpathingsolver/types.ts
@@ -92,7 +92,7 @@ export interface HgPortPointPathingSolverParams {
   colorMap?: Record<string, string>
   inputSolvedRoutes?: SolvedRoutesHg[]
   layerCount: number
-  effort: number
+  effort?: number
   preserveTerminalPcbPortIds?: boolean
   minViaPadDiameter?: number
   flags: {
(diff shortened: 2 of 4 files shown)


thairc-dev:
Unit test `tests/get-effort-scale.test.ts` passes 100% (5/5). Fixed TypeScript types for HgPortPointPathingSolverClass and TinyHypergraphPortPointPathingSolver. All CI checks are green (11/11). Ready for review!

thairc-dev:
Friendly bump on this PR. All CI checks pass 100% and unit tests are included.

PR #1754, opened 2026-07-23
Fix Pipeline7 routing for srj24 sample 4

ShiboSoftwareDev:
## Summary

- detect complete pad grids nested inside mixed-pitch components and route them with component-local BGA topology
- add legal target/free-region via access and connect those overlapping regions in the capacity graph
- use deterministic large-graph route ordering so cross-layer routes are handled before same-layer routes within each net
- cache the solved pathing output used to prepare high-density routing
- bound best-effort global and exact DRC finishing work only on boards with at least 500 routes

## Root cause

The reported route was genuinely unreachable: the mixed-pitch array package was not recognized as a BGA, so its top and bottom terminals had no legal layer-transition path in the global capacity graph.

After repairing that topology, sample 4 exposed two large-board performance costs: Pipeline7 repeatedly rebuilt and linearly searched the full pathing output while constructing high-density inputs, and its best-effort DRC finishing stages exceeded the benchmark deadline.

## Step visualization

```mermaid
flowchart LR
  A["Mixed-pitch component"] --> B["Extract nested BGA grid"]
  B --> C["Create legal cross-layer access"]
  C --> D["Prioritize large-net via routes"]
  D --> E["Reuse prepared pathing output"]
  E --> F["Bound best-effort exact DRC"]
  F --> G["srj24 sample 4 solved"]
```

Every behavioral fix has a checked-in SVG snapshot generated from the same inputs and outputs asserted by its test.

### 1. Mixed component → component-local BGA

The complete 5×5 grid is selected from the mixed component, while the unrelated perimeter pads remain global.

Nested BGA topology before, detection step, and after

### 2. Dense component → legal via-access graph

The compact-component gate, layer promotion, and resulting capacity edge are shown in sequence. The test exercises both capacity-edge implementations.

Dense-component detection and via-access before and after

### 3. Input order/repeated preparation → prioritized and cached pathing

The snapshot shows the cross-layer route moving to the front of the larger net, the large-graph heuristic weight, and both `getOutput()` calls reusing one indexed result.

Large-graph ordering and output caching before and after

### 4. Unconditional exact DRC → bounded large-board finish

For the 500-route fixture, the estimated 5,250,000 checks exceed the 5M guard. The bounded solver preserves the routes without invoking the exact DRC evaluator; the small-board path still runs normally.

Exact DRC finishing before and after

## Validation

- `bun run build`
- `bun run format:check`
- four focused regression tests, each with a before/after SVG snapshot
- [`srj24` sample 4 benchmark run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/30101380246)
  - completed: 100% (1/1)
  - timed out: 0/1
  - sample 4: solved in 1254.0s with 669 vias
  - relaxed DRC: failed, unchanged from the expected dataset result

This replaces the earlier broad version of the PR. It does not include the route-stitching changes or snapshot churn from that version.


ShiboSoftwareDev:
Regression follow-up pushed in b0ddbeb.

Final-code benchmark verification:
- srj24 sample 2: solved
- srj18 sample 3: solved
- srj18 sample 6: solved
- srj24 sample 4: still solved

Local verification:
- 13 focused regression/topology tests pass
- bunx tsc --noEmit passes
- bun run build passes
- git diff --check passes

The remaining broad-suite failures were reproduced unchanged at the original PR commit 42850be: bugreport51 (12.124% snapshot delta), bugreport57 (45.552% snapshot delta), and the pipeline7 SVG-frames timeout with an exact 0.000% snapshot. The existing format check also reports only the same nine pre-existing files, none introduced by this regression-fix commit.

ShiboSoftwareDev:
/benchmark --dataset 24 --sample-numbers 4 --sample-timeout 2000s

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -92,6 +90,47 @@ type PipelineStep<T extends new (...args: any[]) => BaseSolver> = {
   onSolved?: (instance: AutoroutingPipelineSolver7_MultiGraph) => void
 }
 
+const MIN_DENSE_COMPONENT_PAD_COUNT = 90
+const MAX_DENSE_COMPONENT_ASPECT_RATIO = 1.5
+
+export function getCompactDenseComponentBounds(
```
seveibar, line 96: docstring? What does this do? 
```suggestion
/** Merge components into a single obstacle when the number of pads is dense **/
export function getCompactDenseComponentBounds(
```

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -383,7 +434,7 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
     definePipelineStep(
       "edgeSolver",
       CapacityMeshEdgeSolver2_NodeTreeOptimization,
-      (cms) => [cms.capacityNodes!],
+      (cms) => [cms.capacityNodes!, cms.viaDiameter],
```
seveibar, line 437: it would be better to switch to named params- even if you only do it for the second parameter for backwards compat

seveibar: i.e. 
```suggestion
      (cms) => [cms.capacityNodes!, {viaDiameter: cms.viaDiameter }],
```

seveibar: but ideally
```suggestion
      (cms) => [{capacityNodes: cms.capacityNodes!, viaDiameter: cms.viaDiameter }],
```

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/BoundedGlobalDrcBranchPortfolioSolver.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  GlobalDrcBranchPortfolioSolver,
+  type GlobalDrcBranchPortfolioSolverParams,
+} from "high-density-repair03/lib"
+
+const LARGE_BOARD_MIN_ROUTE_COUNT = 500
```
seveibar, line 6: ? don't like this hard coded stat, it seems like a hack that makes things harder to debug

lib/solvers/NodeDimensionSubdivisionSolver/add-target-via-access-layers.ts
```diff
@@ -0,0 +1,164 @@
...
+    contained.center.y - contained.height / 2 + EPSILON &&
+  container.center.y + container.height / 2 >=
+    contained.center.y + contained.height / 2 - EPSILON
+
+const shareConnectionAlias = (
+  a: CapacityMeshNode,
+  b: CapacityMeshNode,
+): boolean => {
+  const aliases = new Set(a._connectedTo ?? [])
+  return (b._connectedTo ?? []).some((alias) => aliases.has(alias))
```
seveibar, line 36: shouldn't need this

seveibar: use connMap (anything else will cause bugs)

PR #1737, opened 2026-07-22
Pin bounded exact DRC repair for SRJ18

AnasSarkiz:
## What changed

- pin `high-density-repair03` to the exact commit from tscircuit/high-density-repair03#30
- refresh only the route snapshots changed by the exact DRC repair
- update the `bugreport-b5b3b9d8` DRC fixture from two trace errors to zero
- bound exact trace-detour variant search when a DRC error does not resolve to a trace pair

## Why

SRJ18 sample 10 reached the repair stage with three exact DRC errors that the existing repair portfolio could not remove. The paired repair PR adds same-net via clearance repair, targeted trace-pair moves/detours, and bounded search for large boards.

The first benchmark-all run exposed a runtime regression: unresolved exact trace-pair errors could cycle through detour variants indefinitely because inapplicable variants did not consume the candidate budget. That loop is now limited to one pass over the finite variant set.

## Impact

- SRJ18 clean samples improve from 2/16 to 3/16 (12.5% to 18.75%)
- sample 10 improves from 3 final DRC errors to 0
- previously clean samples 1 and 5 remain clean
- DRC14 improves from 37/44 to 41/44 clean routes with no previously clean sample regressing
- the eight exact-stage timeout regressions reproduced from dataset01 and SRJ23 now finish in 0.1–8.0 seconds with zero timeouts

## Validation

- `bun run build`
- 15/15 affected autorouter tests pass across 13 files
- 31/31 high-density-repair03 tests pass
- high-density-repair03 typecheck passes
- focused dataset01 timeout rerun: samples 49 and 57 complete in 2.6s and 5.1s
- focused SRJ23 timeout rerun: samples 23, 37, 51, 73, 89, and 106 complete in 0.1–8.0s
- focused SRJ18 run with `--sample-timeout 2000s`: samples 1, 5, and 10 are 3/3 DRC-clean with zero timeouts
- DRC14: 41/44 clean, 73/76 DRC issues removed

Depends on: https://github.com/tscircuit/high-density-repair03/pull/30

PR #1756, opened 2026-07-23
Fix focused snapshot updates with selectable CI runner

AnasSarkiz:
## Summary

- keep `/usf` on the configured fast benchmark runner by default
- add `/usf --ubuntu-latest` for GitHub-hosted x64 snapshot parity
- pin Bun 1.3.8 and regenerate the exact failed CI test shard in either mode
- restrict focused commits to snapshots belonging to failed tests
- check out the exact PR head and refuse stale pushes
- include the source run, shard, runner mode, architecture, and changed snapshots in the bot comment

## Motivation

PR #1737 fails SVG snapshots for `bugreport51-7db9f8` and `bugreport58-b69d72` in Bun Test shard 2. The fast ARM runner may not reproduce x64-only snapshot differences, so focused updates now support an explicit authoritative retry:

- `/usf` — configured benchmark runner, falling back to `blacksmith-32vcpu-ubuntu-2404-arm`
- `/usf --ubuntu-latest` — GitHub-hosted x64 runner

## Validation

- `git diff --check`
- both workflow YAML files parse successfully
- `bun run build`

Follow-up: after this merges, run `/usf --ubuntu-latest` on #1737 and verify the complete Bun Test matrix.

### DIFF
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -27,7 +27,7 @@ jobs:
         github.event.comment.author_association == 'MEMBER' ||
         github.event.comment.author_association == 'COLLABORATOR'
       )
-    runs-on: ${{ vars.BENCHMARK_RUNNER || 'blacksmith-32vcpu-ubuntu-2404-arm' }}
+    runs-on: ubuntu-latest
     timeout-minutes: 60
 
     steps:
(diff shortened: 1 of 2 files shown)


0hmX (commented):
I do not think this is the correct fix we need to find the code path that caused the change the changes were quite extreme because we only do red for top layer and blue for bottom layer but we are seeing that's some of the obstacle layer change somehow like they went from red to blue that does not make sense.

Operating system should not affect this. Or I am understanding the issue wrong I don't know

AnasSarkiz:
@0hmX Added `/usf --ubuntu-lates` option

.github/workflows/update-snapshots.yml
```diff
@@ -68,25 +72,35 @@ jobs:
...
         uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: 1.3.8
 
       - name: Install dependencies
         if: steps.parse.outputs.can_push == 'true'
         run: bun install
 
+      - name: Install bun-test-plan
```
0hmX: i think this is not needed at least for just snapshot update

.github/workflows/update-snapshots.yml
```diff
@@ -68,25 +72,35 @@ jobs:
...
+        if: steps.parse.outputs.can_push == 'true' && steps.parse.outputs.focused == 'true'
+        run: npm install -g @tscircuit/bun-test-plan
+
+      - name: Generate test plans
```
0hmX: not sure really if we need this

0hmX (approved):

PR #1758, opened 2026-07-23
Render multi-layer obstacles with red opacity

0hmX:
## Summary

- render each logical obstacle exactly once instead of once per z-layer
- keep bottom-only obstacles blue at 50% opacity
- render top-only and multi-layer obstacles red, with opacity increasing by canonical layer count: `1 - 0.5^N`
- preserve compound layer metadata such as `z0,1`

This is a separate implementation based directly on `main`. It replaces the neutral-gray multi-layer treatment from #1757 with the requested layer-weighted red opacity and does not alter #1737 or #1757.

## Why the color appeared to change

The obstacle's routing layers were not being mutated. The renderer emitted coincident rectangles for every z-layer, so semi-transparent colors accumulated at the same coordinates and the final visible color depended on layer/render order. This computes the equivalent red opacity on one rectangle instead.

Examples:
- 1 top layer: red at 50%
- 2 layers: red at 75%
- 3 layers: red at 87.5%
- bottom-only: blue at 50%

## Validation

- `bun test tests/utils/convertSrjToGraphicsObject-layer-colors.test.ts tests/bugs/bugreport59-82431e.test.ts --timeout 9999999`
- `bun run build`
- `git diff --check`
- inspected the generated bugreport59 PNG
- verified all 112 route polylines in the regression snapshot are unchanged

Related discussion: https://github.com/tscircuit/tscircuit-autorouter/pull/1737#discussion_r3636174280


0hmX:
@copilot help conflicts and update snapshots

PR #1763, opened 2026-07-24
Fix Pipeline7 routing for SRJ24 sample 4

ShiboSoftwareDev:
## What changed

- Detect a complete nested BGA pad grid inside a mixed component and plan that grid as component-local topology.
- Preserve physically valid, via-sized cross-layer access between dense-component target and free capacity regions.
- For hypergraphs with at least 500 connections, route cross-layer pairs first within each net and use a stronger A* heuristic.

## Root cause

SRJ24 sample 4 contains a complete 13x13 BGA grid inside a 337-pad mixed component. Whole-component BGA detection rejects that component because of its additional perimeter pads, so the global topology reserved the BGA area without generating a legal component-local escape topology. After recognizing the nested grid, dense target/free regions still retained disjoint layer availability even where a via physically fits. Once those topology problems were fixed, the resulting 841-route hypergraph needed cross-layer routes prioritized to finish within the existing effort-1 iteration cap.

## Scope and impact

- Nested-grid detection requires a complete rectangular grid with at least 5 positions per axis.
- Via-access promotion is restricted to undetected compact components with at least 90 pads and a maximum 1.5 aspect ratio, and only applies where a via-sized physical overlap exists.
- The routing policy change applies only to graphs with at least 500 connections.
- No terminal normalization, route stitching, or high-density solver behavior is changed.
- No snapshots are changed.

Compared with the superseded broad approach, this PR is reduced from 42 files / 4,264 additions to 12 files / 719 additions.

## Verification

- `bun run build`
- `bun test tests/features/topology/nested-bga-topology.test.ts tests/features/topology/target-via-access-layers.test.ts tests/features/topology/large-graph-connection-order.test.ts tests/node-dimension-subdivision-solver.test.ts tests/node-dimension-subdivision-connectivity-bridge.test.ts --timeout 9999999`
- Pipeline7, SRJ24 sample 4, effort 1: `Success: yes` on latest main; pathing, high-density routing, stitching, and exact DRC stages completed.
- Pipeline7, SRJ24 sample 2, effort 1: `Success: yes`; the former duplicate `pcb_port_118` stitching failure is absent.
- Pipeline7, SRJ18 sample 3, effort 1: `Success: yes`; the former unknown `pcb_port_50` stitching failure is absent.
- Pipeline7, SRJ18 sample 6, effort 1: pathing and stitching complete; the former large-problem iteration exhaustion is absent. The full runner also completed successfully before rebasing onto main's newly pinned exact-DRC repair.

Relaxed DRC remains failing for these benchmark outputs, as it does for the other solved SRJ24 scenarios; this PR fixes solver completion, not DRC quality.

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
@@ -92,6 +92,50 @@ type PipelineStep<T extends new (...args: any[]) => BaseSolver> = {
   onSolved?: (instance: AutoroutingPipelineSolver7_MultiGraph) => void
 }
 
+const MIN_DENSE_COMPONENT_PAD_COUNT = 90
+const MAX_DENSE_COMPONENT_ASPECT_RATIO = 1.5
+
+export function getCompactDenseComponentBounds(
+  obstacles: SimpleRouteJson["obstacles"],
+  detectedComponentIds: ReadonlySet<string>,
+): SimpleRouteJson["bounds"][] {
+  const obstaclesByComponentId = Map.groupBy(
+    obstacles.filter(
+      (obstacle) =>
+        obstacle.componentId &&
+        !detectedComponentIds.has(obstacle.componentId),
+    ),
+    (obstacle) => obstacle.componentId!,
+  )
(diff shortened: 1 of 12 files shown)


ShiboSoftwareDev:
Superseded by #1754, which now contains the focused implementation rebased onto current main. srj24 sample 4 was verified locally with the official 2000s benchmark command and solved in 1211.2s.

PR #1768, opened 2026-07-24
Add bus types to SimpleRouteJson

seveibar:
## Summary

- make `lib/types/srj-types.ts` the canonical public source for the SimpleRouteJson type family
- stop re-exporting or intersecting SimpleRouteJson and SimplifiedPcbTrace from `high-density-repair03`
- preserve the previously exposed SRJ surface locally, including `zLayers`, connection root metadata, via-to-pad clearance, and terminal wire port IDs
- normalize public `zLayers` into the autorouter's internal `__zLayers` representation
- add a `SimpleRouteBus` type with `busId`, ordered `connectionNames`, and optional `maxLengthSkew`
- verify public bus metadata survives preprocessing

```ts
buses: [
  {
    busId: "data",
    connectionNames: ["data0", "data1", "data2"],
    maxLengthSkew: 0.1,
  },
]
```

## Ownership

`@tscircuit/capacity-autorouter` now defines and exports its own SRJ types. `high-density-repair03` remains an implementation dependency for DRC repair, but it no longer supplies the autorouter package's public SimpleRouteJson definition.

## Scope

This PR proposes the SRJ representation only. It does not add bus length-matching behavior, change routing, or alter via-in-pad defaults.

## Validation

- `bun run build`
- `bun run format:check`
- `bun test tests/features/simple-route-json-bus-types.test.ts tests/utils/layer-count-normalization.test.ts`

All checks pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ interface SimpleRouteJson {
   minTraceWidth: number
   obstacles: Obstacle[]
   connections: Array<SimpleRouteConnection>
+  buses?: Array<SimpleRouteBus>
   bounds: { minX: number; maxX: number; minY: number; maxY: number }
   traces?: SimplifiedPcbTraces // Optional for input
 }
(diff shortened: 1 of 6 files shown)


seveibar:
CC @0hmX

PR #1776, opened 2026-07-28
fix: maxInnerIterationsPerGrowthAttempt was silently ignored

apullin:
`GrowShrinkHighDensityIntraNodeSolver` bounds a single growth attempt by
assigning `activeSubSolver.MAX_ITERATIONS`. `PortfolioSingleIntraNodeSolver`
then recomputes `MAX_ITERATIONS` in `refreshDynamicIterationLimit()` from the
remaining candidate budgets — on its first step, and again on every portfolio
expansion — so the cap was discarded immediately. Observed on the added
fixture: a cap of 25 becomes 45,128.

The `maxInnerIterationsPerGrowthAttempt` parameter has therefore never had any
effect on this path, and a growth attempt always ran to the portfolio's own
dynamic budget before the node was grown and retried.

**Fix:** an explicit `externalMaxIterations` ceiling on the portfolio that the
dynamic limit respects, set by GrowShrink alongside the existing assignment.
No production caller currently sets the parameter (neither pipeline supplies
it), so default behavior is byte-identical; the fix only makes the documented
parameter actually work for callers who pass it.

**Tests:** `tests/features/never-fail-growth-high-density/caps-inner-iterations.test.ts`
covers both the wiring (a busy 4-connection node whose inner solver is still
running after one step) and the portfolio ceiling itself. Both fail before this
change and pass after.

Suite: 419 pass / 55 skip / 2 fail; the two failures (`bugreport36-d4c6c2` and
`dip16 crossing traces 1206x4`) reproduce identically on an unmodified checkout
of this commit's parent (v0.0.717), so they are pre-existing and unrelated.

### DIFF
--- a/lib/solvers/HyperHighDensitySolver/GrowShrinkHighDensityIntraNodeSolver/GrowShrinkHighDensityIntraNodeSolver.ts
+++ b/lib/solvers/HyperHighDensitySolver/GrowShrinkHighDensityIntraNodeSolver/GrowShrinkHighDensityIntraNodeSolver.ts
@@ -148,6 +148,10 @@ export class GrowShrinkHighDensityIntraNodeSolver extends BaseSolver {
       ),
     })
     if (this.constructorParams.maxInnerIterationsPerGrowthAttempt) {
+      // Must be an external ceiling: the portfolio recomputes MAX_ITERATIONS
+      // dynamically on its first step and would otherwise discard this.
+      this.activeSubSolver.externalMaxIterations =
+        this.constructorParams.maxInnerIterationsPerGrowthAttempt
       this.activeSubSolver.MAX_ITERATIONS =
         this.constructorParams.maxInnerIterationsPerGrowthAttempt
     }
--- a/lib/solvers/HyperHighDensitySolver/PortfolioSingleIntraNodeSolver.ts
+++ b/lib/solvers/HyperHighDensitySolver/PortfolioSingleIntraNodeSolver.ts
@@ -279,6 +279,14 @@ export class PortfolioSingleIntraNodeSolver extends HyperParameterSupervisorSolv
     if (typeof setup === "function") setup.call(solver)
   }
 
+  /**
+   * Hard ceiling imposed by an owner, e.g. GrowShrink's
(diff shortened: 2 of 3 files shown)


seveibar:
@apullin can you give us some context on your tscircuit usage? Will help us understand if the ai generated PRs are based in real usage (which validates them more)

apullin:
@seveibar 
They are based in real usage. The 3 PR's I just put up are bugs that came out of a large campaign that speeds up autorouting by 4-5x.

seveibar:
CC @ShiboSoftwareDev  for review

@apullin we will merge more quickly w/ context that helps build trust in the contributor/maintainer. Are you currently building a board with tscircuit tsx? Or is the autorouter used standalone? In general, what is the complexity of the board you are routing in terms of net count, or dominant chip (e.g. RK3566, RP2040 etc.)

If you're just here for optimization, that's also ok- but we will need to have a maintainer help to make sure your changes get upstream through the pipeline (often optimizers have a more narrow outlook on the objectives, and don't necessarily see the upstream visual snapshots)

We can also help share insights in either case, e.g. we ran experiments on high density parallelism as well as high density lossy caching

seveibar (commented):
I don't think i can review because this is @ShiboSoftwareDev 's system, it looks OK to me, there may be a small negative performance impact based on the benchmark results

seveibar:
the perf impact is likely to be negligible for this one in particular- but this could just be a set up for a future optimization

apullin:
@seveibar No boards started yet. Will do some S100 boards eventually. Only looking at speedups on existing benchmarks.

seveibar:
cool- it can be a useful way to report on optimization results if you PR a bugreport for a board you're trying to route- https://docs.tscircuit.com/contributing/report-autorouter-bugs

also note that there's a lot of discussion on discord (https://tscircuit.com/join) in the #autorouting channel about various approaches and optimizations. We're currently more focused on predefined trace influence (for staged autorouting), power traces, ddr3 routing etc. than optimization. That said, a good optimization PR is appreciated especially if it doesn't come at the cost of code clarity

glad to have you here, normally reviews go pretty quick but i will leave this particular subsystem to @ShiboSoftwareDev

apullin:
alright, leave the bug in I guess

PR #1799, opened 2026-07-30
fix(topology): detect BGA cores in mixed packages

ShiboSoftwareDev:
## Stack

Stacked on the test-only reproduction PR #1798.

## Problem

Sample 4 has a regular BGA grid in the middle of a package with different-shaped pads around it. All those pads share the same component ID.

The detector checked every pad together, so the different outer pads prevented it from recognizing the BGA grid.

## Fix

1. Find the regular grid of same-sized, evenly spaced pads inside the package.
2. Treat only that grid as BGA-local. Keep the surrounding pads in the normal global topology.

The existing BGA topology solver can then create the missing connection from the top layer to the bottom layer.

| Reproduction | Fixed |
| --- | --- |
| before | after |

Green pads are the detected BGA core. Blue pads stay global. The last panel shows the BGA-local layer connection.

The fixed path contains more graph nodes, so the SVG table automatically makes its cells wider. The pad coordinates and spacing are unchanged.

## Sample 4 result

Pipeline7 now gets past the original static reachability failure and completes stages 1–10.

It later reaches a separate path-search iteration limit on current `main`. This PR fixes the missing topology, but does not claim that the full sample currently solves.


seveibar (commented):
Shouldn’t this be a different detector?

seveibar (approved):

PR #1809, opened 2026-07-31
Use B01 with regional reroute fallback in Pipeline9

seveibar:
## What changed

- Adds Pipeline9 as a copied, independently staged pipeline for preloaded-trace routing.
- Loads pre-routed copper into the tiny hypergraph as stable initial assignments without changing hypergraph topology or adding approximation rectangles.
- Routes high-density regions around immutable preloaded copper with `@tscircuit/high-density-b01`.
- If B01 cannot solve a region, reroutes all copper clipped by that region with the regular high-density portfolio and splices only those replacements back into the original traces.
- Runs relaxed joint DRC against preloaded and newly routed copper, then repairs conflicts by selectively making affected preloaded traces movable.
- Resolves same-net via reuse from connection metadata, without geometric connectivity inference.
- Marks intentional preloaded-trace mutations with explicit replacement metadata; a colliding `pcb_trace_id` alone does not remove preloaded copper.

## Why

Pipeline9 needs to preserve already-routed subcircuits as real connectivity and copper, route around them first, and mutate them only when a local region is otherwise unsolvable. Explicit replacement metadata also keeps the shared relaxed-DRC evaluator correct for other pipelines whose generated trace ids may collide.

## Validation

- Pipeline9 focused/regression tests, including SRJ23 samples 3, 4, 10, 21, 25, 37, 44, 46, 50, 58, 65, 70, and 74
- Pipeline9 solved visual snapshots for SRJ23 samples 1, 3, and 10, with preloaded traces visible
- Pipeline7 SRJ27 power-expansion DRC regression
- `bun run format:check`
- `bunx tsc --noEmit`
- `bun run build`
- Latest completed SRJ23 benchmark: 100.0% completion, 97.4% relaxed DRC, 0/76 timeouts

Samples 42 and 72 are the only relaxed-DRC failures; their corrected inputs contain unavoidable different-net terminal-pad overlap.

## Related

The exact B01 failure input is isolated separately in tscircuit/high-density-b01#6.


seveibar:
Superseded by the reviewable stack below:

1. #1954 — explicit initial connectivity
2. #1955 — immutable via anchors during simplification
3. #1956 — B01 routing around fixed preloaded copper
4. #1957 — explicit regional reroute fallback and trace replacement
5. #1958 — joint DRC accounting with preloaded copper
6. #1959 — selective repair of implicated preloaded traces
7. #1960 — regional B01 joint-D﻿RC repair
8. #1961 — terminal escape relocation within the owning pad
9. #1962 — solved SRJ23 visual snapshots (including sample 3)

The stack is based on current main (`d116faf1`). The ineffective targeted-dogleg postpass from this monolith was intentionally not carried over: it did not improve either residual relaxed-D﻿RC case. The benchmark is running on #1961; #1962 is visual-only.

seveibar:
Stack update: #1963 (bounded preliminary joint-D﻿RC search) is inserted between #1961 and the visual-only #1962. It removes the corrected SRJ23 sample 52/53 timeouts while preserving 100.0% completion / 97.4% relaxed DRC locally.

PR #1819, opened 2026-07-31
fix(topology): align cross-layer target access

ShiboSoftwareDev:
Stacked on #1812, which adds the reproduction and shared SVG fixture.

## Problem

The global topology owns a target footprint on one layer. Inside the same footprint, component topology already describes a same-net target on another layer and free capacity on the layers between them.

Topology merging did not combine those descriptions, so it produced aligned single-layer slices with no node that could change layer. That is why `srj24` sample 4 failed its static reachability check.

## Fix

Before merging, this change checks whether the existing global target has uniform cross-layer access:

1. Require an overlapping same-net component target on another layer.
2. Divide the target footprint at the existing topology boundaries for validation.
3. Require every part of the footprint to expose the same contiguous layers.
4. Reject the change if any other net blocks the footprint or if the target is smaller than the configured via.
5. If all checks pass, add only the actual same-net target layers to the ordinary target node's `availableZ` and `layer` fields. Intermediate free layers prove via clearance but do not become target-owned.

The target is not split and no new region type is introduced. There is no portal flag, special edge, pipeline stage, or pathfinder exception. The existing topology merger and edge solver handle the resulting ordinary multi-layer node.

The restoration step also keeps the global target authoritative instead of restoring an overlapping component copy on top of it.

## Snapshot

Uniform cross-layer target access

This is the exact same fixture as #1812:

1. **Physical copper input** is unchanged.
2. **Topology inputs** are unchanged. Blue regions are component-local free capacity, not copper.
3. **Merged capacity output** now contains one normal target region on z0 and z2. The z1 row stays blue/free; it only proves that a via can pass through the footprint. The green line is the resulting graph path between the purple terminals.

## Verification

- `bun run build`
- `bun run format:check`
- `bun test --timeout 9999999 tests/features/topology/same-net-cross-layer-targets.test.ts tests/features/topology/topology-merging`
- Local sample 4 topology diagnostic: former target `cmn_1840` is one ordinary node on z0 and z5, the intervening layers remain free, and the static reachability precheck passes.
- Hosted benchmark (pending): `/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1`


ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 3000s --concurrency 1

PR #1820, opened 2026-08-01
test(pathing): visualize large-graph prepass cutoff

ShiboSoftwareDev:
Stacked on #1819, which restores sample 4's missing cross-layer topology path.

## Problem

Before the main tiny-hypergraph solve, `DuplicateCongestedPortSolver` can independently trace each connection, find boundary ports that many routes want to use, and add nearby alternatives.

That prepass currently runs only when a graph has at most 180 connections. `srj24` sample 4 has 841 connections, so the prepass is skipped. The main selective-rerip solve then repeatedly rearranges congested routes and runs out of iterations.

## Reproduction

Large-graph prepass cutoff

The orange line is the production connection-count cutoff. The two points call the same production predicate as the solver: a 100-connection graph runs the prepass, while sample 4 is red because 841 is above 180.

This PR only extracts the existing predicate and constant so the behavior can be tested and visualized. It does not change which graphs run the prepass.

The fix is stacked in #1821 and uses this exact test source unchanged.

## Verification

- `bun test tests/features/port-point-pathing/duplicate-congested-port-prepass-cutoff.test.ts`
- `bun run build`
- `bun run format:check`


ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 18 --sample-numbers 6 --effort 1 --sample-timeout 3000s --concurrency 1

Issue #140, opened 2025-06-10
Autorouting Bug Report: pcb_trace overlap error


techmannih:
link
https://tscircuit.com/editor?package_id=57122ccc-9eb2-48cb-a381-c42fdd82838c&file_path=index.tsx

PR #1846, opened 2026-08-02
test: reproduce sample 4 pathing crossings

ShiboSoftwareDev:
## Stack

Stacked on #1845, which gives duplicated graph ports real physical lane coordinates.

## Problem

Port-point pathing must not connect two different nets as crossing chords inside a node that has only one available layer. There is no second copper layer inside that node where one route can pass over the other.

The earlier version of this PR compared pathing with uniform port distribution. That was misleading: the crossing already exists in the pathing output.

## Reproduction

The test runs srj24 sample 4 only through port-point pathing and checks every route pair in every single-layer node before uniform distribution or high-density routing runs.

The hosted result finds 382 different-net crossings. The first is in cmn_188, which has only z0. All four endpoints report zero distance from the real node boundary, so this is real invalid pathing topology rather than a drawing approximation.

[Hosted reproduction run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/30913283091)

## Root cause and fix stack

The same bug has a small native tiny-hypergraph reproduction and fix:

- [tiny-hypergraph #139](https://github.com/tscircuit/tiny-hypergraph/pull/139) shows the timeout-only greedy fallback committing a crossing.
- [tiny-hypergraph #140](https://github.com/tscircuit/tiny-hypergraph/pull/140) keeps hard crossing rejection in greedy routing and validates the final endpoint hop.

Those PRs use one unchanged solver fixture and its native SVG snapshot. Before the fix, both crossing routes are committed as solid lines. After the fix, the invalid route stays uncommitted as a dashed request.

## Validation

- GitHub Actions only; nothing was run locally.
- This reproduction fails with Received: 382, Expected: 0.
- The upstream fix suite passes on GitHub Actions.

### DIFF
--- a/tests/features/srj24-sample4-boundary-order-handoff.test.ts
+++ b/tests/features/srj24-sample4-boundary-order-handoff.test.ts
@@ -93,6 +93,9 @@ test(
 
     const afterUniform = solver.uniformPortDistributionSolver!.getOutput()
     const afterCrossings = findDifferentNetSingleLayerCrossings(afterUniform)
+    const affectedNodeIds = new Set(
+      [...beforeCrossings, ...afterCrossings].map(({ nodeId }) => nodeId),
+    )
 
     console.log(
       "srj24 sample 4 boundary-order handoff",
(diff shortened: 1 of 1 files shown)


ShiboSoftwareDev:
Superseded by the upstream hard-constraint reproduction/fix (tscircuit/tiny-hypergraph#139 and #140), integration PR #1847, and the focused remaining search-stall reproduction #1848. The hosted diagnostic here established that crossings already existed before uniform distribution; #1847 now rejects those invalid crossings instead of passing them downstream.

Issue #142, opened 2025-06-15
Autorouting Bug Report: <board#470 /> seveibar/led-water-accelerometer


seveibar:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=857612b1-a0a4-47d9-9155-9c3855a32339

![Image](https://github.com/user-attachments/assets/4000c3cb-64b1-4ea9-b3e5-b3e1baebd83d)

ricohageman:
@seveibar this board routes completely, can you close it? <img width="452" height="596" alt="Image" src="https://github.com/user-attachments/assets/0023b538-3d00-4a42-a2a7-dfce3b78dc09" />

Issue #141, opened 2025-06-11
Autorouting Bug Report: <board#294 />


imrishabh18:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=6c547ec5-ce7f-4145-a2db-2c1e2440a896

(any additional details? Maybe a screenshot?)

PR #1849, opened 2026-08-02
Avoid duplicate pathing ports within one net

ShiboSoftwareDev:
## Problem

The duplicate-port prepass sizes boundary lanes by route count. A net can have several MST route branches, but the pathing solver permits those branches to share a port because they are electrically the same net.

## Fix

Pin the upstream `tiny-hypergraph` correction:

- Reproduction: [tiny-hypergraph #141](https://github.com/tscircuit/tiny-hypergraph/pull/141)
- Fix: [tiny-hypergraph #142](https://github.com/tscircuit/tiny-hypergraph/pull/142)

The prepass now counts distinct electrical nets at each port. Different nets still receive separate physical lanes; branches of one net share one lane.

This integration changes only the exact dependency SHA. It adds no sample-specific threshold or fallback.

## Visualization

The reproduction and fix use the same real solver fixture and snapshot path.

| Before: 2 ports for one net | After: 1 shared port |
| --- | --- |
| Before | After |

The upper frame is the one-port input. The lower frame is the actual duplicate-prepass output. Gray circles are topology ports; colored dashed lines are same-net route endpoint hints.

## Hosted validation

Upstream CI passes 109/109 tests.

The exact benchmark was run in GitHub Actions:

```text
/benchmark-long --dataset srj24 --sample-numbers 4 --sample-timeout 2000s --concurrency 1
```

[Hosted benchmark result](https://github.com/tscircuit/tscircuit-autorouter/pull/1849#issuecomment-5157716670): sample 4 still fails safely in selective-rerip pathing after 318.6 s. The correction reduces the graph from 76,634 to 76,334 ports, but the same route remains blocked by a separate missing local cross-layer region.

This PR is therefore a valid general correctness fix, but it is not presented as the sample 4 fix.

## Stack

Stacked on #1847. The focused sample 4 topology follow-up is separate so this dependency pin stays independently reviewable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "recharts": "^2.15.1",
     "tailwind-merge": "^3.2.0",
     "terser": "^5.43.1",
-    "tiny-hypergraph": "git+https://github.com/tscircuit/tiny-hypergraph.git#2ac358656476d25c4a938a75cac464b499d1bdbf",
+    "tiny-hypergraph": "git+https://github.com/tscircuit/tiny-hypergraph.git#058e5189b13f7832235e72281dc27c006cafeb71",
     "tiny-hypergraph-poly": "git+https://github.com/tscircuit/tiny-hypergraph.git#7b93b4c",
     "tsup": "^8.3.6",
     "typescript": "^5.9.3",


ShiboSoftwareDev:
/benchmark-long --dataset srj24 --sample-numbers 4 --sample-timeout 2000s --concurrency 1

PR #1851, opened 2026-08-02
fix(topology): preserve BGA target root aliases

ShiboSoftwareDev:
Stacked on #1850.

## Problem

The route already says that `source_trace_3` and `source_net_3` are two roots of the same electrical connection. BGA topology generation was keeping only `source_trace_3` on each generated target node.

The overlapping global target is identified by `source_net_3`. Because the BGA node had dropped that root ID, the existing topology merger could not recognize the targets as the same net and did not create the legal local layer transition.

## Fix

Keep every `__rootConnectionNames` entry on generated BGA target nodes:

- `_targetConnectionName` remains the first root, preserving existing behavior.
- `_connectedTo` stores all roots, preserving the full electrical identity.

The existing topology merger can then match the BGA and global target through `source_net_3` and build its already-supported regular aligned regions. This adds no new region type, portal, routing fallback, or geometry exception.

## Snapshot

This uses the fixture and test from #1850 unchanged.

Preserved BGA net roots

The third frame reports the root IDs retained by the actual generated target nodes. The fourth frame is the actual layer-access graph; columns are graph regions and do not represent physical X spacing.

## Hosted verification

- GitHub Actions test suite: pending
- `srj24` sample 4, Pipeline 7, effort 1: pending


ShiboSoftwareDev:

PR #1856, opened 2026-08-02
fix(topology): split partial BGA obstacle overlaps

ShiboSoftwareDev:
Stacked on #1855, which contains the unchanged reproduction and SVG fixture.

## Issue

A BGA routing rectangle can be covered by an obstacle on only one layer and only part of its XY area.

The old removal code treated that as if the obstacle covered the entire rectangle. It removed the blocked layer everywhere and emitted one node per remaining layer. In sample 4, that turns one regular BGA gap into five isolated rectangles and discards the uncovered z5 sliver.

## Fix

Split the routing rectangle at the real obstacle boundary:

- pieces outside the obstacle keep the original layer set
- the overlapping piece keeps all layers except the obstacle layers
- each piece remains an ordinary CapacityMeshNode
- remaining layers stay grouped in one node instead of being copied into one node per layer

This models target-only, target-on-one-layer/free-on-other-layers, and free-only XY regions directly. It adds no portal, via region, via-in-pad shortcut, special edge, pathfinder exception, or pipeline stage.

## Visualization

The snapshot is the exact same test and fixture as #1855. Horizontal position and width are the real board x geometry; rows are z0-z5. Red is target capacity, blue is free capacity, and a vertical line means one ordinary node spans multiple layers.

Before the fix, the covered gap is five isolated z0-z4 nodes and z5 is removed everywhere. After the fix, the covered part is one z0-z4 node and the small uncovered part remains one z0-z5 node.

## Hosted validation

All tests, snapshots, and the srj24 sample 4 benchmark run in GitHub Actions. Nothing is executed locally.

Benchmark command after the focused snapshot check passes:

### DIFF
--- a/lib/solvers/BgaTopologyGeneratorSolver/RemoveMeshNodeOverlappingSolver.ts
+++ b/lib/solvers/BgaTopologyGeneratorSolver/RemoveMeshNodeOverlappingSolver.ts
@@ -1,17 +1,73 @@
-import {
-  doBoundsOverlap,
-  getBoundFromCenteredRect,
-} from "@tscircuit/math-utils"
+import { getBoundFromCenteredRect, type Bounds } from "@tscircuit/math-utils"
 import { BaseSolver } from "@tscircuit/solver-utils"
 import type { GraphicsObject } from "graphics-debug"
 import type { CapacityMeshNode, Obstacle } from "lib/types"
 import { createRectFromCapacityNode } from "lib/utils/createRectFromCapacityNode"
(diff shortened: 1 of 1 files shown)


ShiboSoftwareDev:

ShiboSoftwareDev:
Superseded by the directed-search fix in #1870. This topology split made sample 4 searchable, but it changed many routes and produced unacceptable intersections/trace geometry. The replacement leaves BGA topology unchanged.

PR #1860, opened 2026-08-02
experiment(pathing): prefer shorter tiny-hypergraph paths

ShiboSoftwareDev:
Temporary hosted-only experiment stacked on #1856. It increases the physical-distance term for solveGraph so A* does not exhaust the local multilayer search around close cross-layer terminals. The sample 4 benchmark will decide whether this direction is valid; this PR is not part of the proposed fix stack unless the hosted result is both solved and clean.


ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted sample 4 validation failed: Pipeline7 exhausted selective-rerip iterations after 487.0s. Increasing the path distance cost does not address the root cause, so this temporary experiment is being closed.

ShiboSoftwareDev:
Closing: hosted benchmark disproved this tuning approach.

PR #1861, opened 2026-08-02
experiment(topology): build point pairs by layer

ShiboSoftwareDev:
## Purpose

Temporary hosted validation for the remaining srj24 sample 4 failure.

The current point-pair topology uses only XY distance. In a staggered top/bottom pad field, it alternates between layers because the opposite-layer pad is closest. That creates many short routes that each require a layer change.

This experiment builds the same-layer trees first, then adds only the cross-layer edges needed to connect them. It is not the final review stack; it will be replaced by a focused visual reproduction and fix if the hosted benchmark passes.

## Validation

No solver or tests are run locally. The sample 4 benchmark and all tests run in GitHub Actions.

### DIFF
--- a/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
+++ b/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
@@ -4,7 +4,7 @@ import {
   SimpleRouteJson,
 } from "lib/types"
 import { BaseSolver } from "../BaseSolver"
-import { buildMinimumSpanningTree } from "./buildMinimumSpanningTree"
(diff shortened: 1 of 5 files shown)


ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted validation disproved this approach. Sample 4 still exhausted iterations after 399.1s, the full test matrix showed broad unrelated routing diffs, and the focused diagnostic showed 154 independent cross-layer nets remain. The remaining stall was a same-layer bottom route, so MST ordering is not the root cause.

ShiboSoftwareDev:
Closing the regression-heavy experiment; it did not solve sample 4.

PR #1862, opened 2026-08-02
experiment(topology): keep narrow BGA fragments single layer

ShiboSoftwareDev:
## Purpose

Temporary hosted validation for the invalid intersections and remaining srj24 sample 4 failure.

The initial BGA topology only makes a region multilayer when both dimensions can fit a via. The partial-overlap split accidentally breaks that invariant: it can create a narrow fragment but retain every layer. The pathfinder may then change layers in a region where the physical solver cannot place a via.

This experiment keeps an overlap fragment multilayer only when it still meets the existing BGA via-fit threshold. Smaller fragments become coincident ordinary single-layer regions, so traces may cross them on each legal layer but may not change layers there.

## Validation

No project code is run locally. GitHub Actions runs the fixture, test matrix, and srj24 sample benchmarks.

### DIFF
--- a/lib/solvers/BgaTopologyGeneratorSolver/BgaTopologyGeneratorSolver.ts
+++ b/lib/solvers/BgaTopologyGeneratorSolver/BgaTopologyGeneratorSolver.ts
@@ -16,6 +16,7 @@ import {
 import type { GraphicsObject } from "graphics-debug"
 import type { CapacityMeshNode, Obstacle } from "lib/types"
 import { createRectFromCapacityNode } from "lib/utils/createRectFromCapacityNode"
+import { getViaDimensions } from "lib/utils/getViaDimensions"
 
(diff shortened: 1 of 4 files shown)


ShiboSoftwareDev:
/benchmark-long --dataset srj24 --sample-numbers 3,4 --sample-timeout 4000s --concurrency 2

ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted validation failed. Sample 3 exhausted pathing after 185.2s and sample 4 after 471.1s/474.4s. Keeping narrow fragments single-layer prevents physically impossible local vias, but by itself does not make the topology routable. This experiment will not be promoted.

ShiboSoftwareDev:
Closing: the hosted benchmark remained unroutable, so this is not the final fix.

PR #1863, opened 2026-08-02
experiment(topology): let topology merging own foreign BGA overlaps

ShiboSoftwareDev:
Temporary hosted-validation PR stacked on #1855.\n\nThis tests one ownership rule: a component-local BGA topology contains only that component's pads. Pads from other components are preserved in their own topology group, and TopologyMergingSolver combines the groups by layer.\n\nThe change removes foreign obstacles from the three BGA-local generation inputs. It does not add fallback routing, synthetic vias, or geometry fragments.\n\nValidation is intentionally GitHub-only. First gate: srj24 sample 4 must solve and pass relaxed DRC. If that succeeds, sample 3 and regression coverage will follow.


ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted validation rejected this experiment. srj24 sample 4 still failed after 541.6s in SelectiveReripTinyHyperGraphSolverWithStableInitialAssignments, and the GitHub test matrix reported broad routing/snapshot regressions. Foreign obstacles cannot simply be removed from every BGA-local topology input. Closing without promotion.

PR #1864, opened 2026-08-02
experiment(topology): keep BGA target nodes component-owned

ShiboSoftwareDev:
Temporary hosted-validation PR stacked on #1855.\n\nThis tests one narrow invariant: the BGA-local generator creates target nodes only for pads owned by that BGA. Foreign pads are still passed to obstacle removal and gap fill, so their copper remains blocked on the correct layer; their authoritative target node comes from their own/global topology group.\n\nProduction diff: one input changes to an empty array. No fallback, synthetic via, or geometry rewrite.\n\nValidation is GitHub-only. The first gate is srj24 sample 4 solved with relaxed DRC pass; ordinary CI must also remain clean.

### DIFF
--- a/lib/solvers/BgaTopologyGeneratorSolver/BgaTopologyGeneratorSolver.ts
+++ b/lib/solvers/BgaTopologyGeneratorSolver/BgaTopologyGeneratorSolver.ts
@@ -41,8 +41,7 @@ export class BgaTopologyGeneratorSolver extends BasePipelineSolver<TopologyGener
               .componentId,
           markedComponentObstacles:
             bgaTopologyGeneratorSolver.markedComponentObstacles,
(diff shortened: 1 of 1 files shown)


ShiboSoftwareDev:

ShiboSoftwareDev:

ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted CI rejected this one-line ownership change. It caused the same broad route changes as the ownership-only experiment and a real regression: srj18 sample 6 failed selective rerip pathing. Foreign target nodes are currently required by downstream topology, so this cannot be promoted. The first sample 4 benchmark was externally cancelled; the re-run was stopped after regression evidence was conclusive.

PR #1866, opened 2026-08-02
experiment(pathing): route layer changes first

ShiboSoftwareDev:
Temporary hosted diagnostic stacked on #1856.

Selective routing already schedules larger nets first, but preserves each net’s original edge order. In srj24 sample 4 that commits 24 same-layer edges before attempting the first constrained cross-layer edge.

This experiment keeps the same net-priority policy and only orders cross-layer edges before same-layer edges within each net. It does not change topology, DRC constraints, via permissions, geometry, or acceptance rules.

Validation is GitHub Actions only; no project command is run locally.

Hosted benchmark:


ShiboSoftwareDev:

ShiboSoftwareDev:
Hosted validation disproved this approach. Sample 4 still exhausted selective-rerip iterations after 384.5s. Prioritizing cross-layer edges is not enough, so this temporary route-order experiment will not be promoted.

ShiboSoftwareDev:
Closing the failed hosted scheduling experiment.

PR #1868, opened 2026-08-02
test(pathing): diagnose remaining layer-aware sample 4 stall

ShiboSoftwareDev:
Temporary hosted-only diagnostic stacked on #1867. It reuses the existing sample 4 pathing diagnostic and reports the active route plus the transition region selected by the layer-aware heuristic after 250,000 iterations. This PR is not part of the fix stack and will be closed after GitHub Actions records the evidence. No project command is run locally.


ShiboSoftwareDev:
Closing the temporary diagnostic PR. Hosted evidence: route 24 still has 24 committed routes after 250k iterations; the heuristic selected distant all-layer region cmn_62 (tiny region 4012), while the candidate tail remained entirely on z0. This confirms the remaining failure is graph guidance toward the transition, not topology legality or an immediate post-transition release.

PR #1870, opened 2026-08-02
fix(pathing): use directed graph guidance for cross-layer routes

ShiboSoftwareDev:
## Change

Pins tscircuit/tiny-hypergraph#146.

For a cross-layer route, the tiny solver now distinguishes a port entering a useful region from the same port entering a dead end. While the route remains on its start layer, A* uses remaining directed graph hops together with the existing XY estimate.

This integration changes only the dependency SHA. It does not include the broad BGA overlap splitting from #1856 and does not alter autorouter topology.

## Validation

- upstream full tests and SVG snapshot: passed in GitHub Actions
- upstream formatter and TypeScript checks: passed in GitHub Actions
- srj24 sample 4 benchmark: pending in this PR

Nothing is run locally.

PR #1871, opened 2026-08-02
test(pathing): diagnose directed sample 4 stall

ShiboSoftwareDev:
Temporary hosted diagnostic stacked on #1870. It stops sample 4 pathing at 250k iterations and reports whether the current route has a directed-hop map, the finite hop range, the best queued candidates, and their path tail. This PR will be closed after the GitHub Actions evidence is collected. Nothing is run locally.


ShiboSoftwareDev:
Hosted evidence collected: at 250k iterations the directed map was present (136,107 reachable directed states, hop range 0–21), but route 25 had a 24,225-candidate queue spread across intermediate layers. Guidance stopped after leaving z0, so the search reverted to XY before reaching its z5 destination. The follow-up fix will retain graph guidance on every non-destination layer.

PR #1873, opened 2026-08-02
experiment(pathing): combine layer-aware point pairs with directed guidance

ShiboSoftwareDev:
## Purpose

Temporary hosted-only validation for srj24 sample 4.

The current point-pair tree uses XY distance alone. In a staggered top/bottom pad field, that creates many avoidable cross-layer connections. The tiny-hypergraph solver then spends most of its iteration budget finding layer transitions instead of routing the board.

This experiment first connects points that share a layer, then adds only the cross-layer edges needed to join those layer groups. It is stacked on the directed tiny-hypergraph guidance in #1870. Neither change relaxes collision checks or accepts invalid geometry.

The diagnostic records connection counts and progress after 100,000 hosted iterations. The acceptance test is the hosted srj24 sample 4 benchmark: the sample must solve, pass relaxed DRC, and produce credible geometry.

No project test or benchmark is run locally.

### DIFF
--- a/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
+++ b/lib/solvers/NetToPointPairsSolver/NetToPointPairsSolver.ts
@@ -4,7 +4,7 @@ import {
   SimpleRouteJson,
 } from "lib/types"
 import { BaseSolver } from "../BaseSolver"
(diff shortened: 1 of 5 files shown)


ShiboSoftwareDev:

PR #1875, opened 2026-08-02
fix(topology): align cross-layer target access in two dimensions

ShiboSoftwareDev:
## Root cause

The BGA topology has one ordinary routing region that crosses a same-net target pad on z5. On z0-z4 that XY area is free; on z5 the target owns it.

The removal step treated the target like a foreign blocker. It removed z5 from the whole region and copied the remaining layers into isolated nodes. Even the first geometric split kept the overlap only on z0-z4, so the closest local region still stopped one layer short of the z5 endpoint.

The hosted diagnostic showed the result directly: the two terminals are 0.37 mm apart, but the pathfinder's best candidates were tens of millimetres away because it had to leave the component to find another layer change.

## Fix

When—and only when—a multilayer BGA region is crossed by a target belonging to the same net as a BGA target:

1. Split the region at the target's existing X or Y boundary.
2. Keep the original layer list on every slice.
3. Let topology merging apply target ownership on the target layer.

This directly represents `[target-only] [target on z5 / free on z0-z4] [free-only]` with ordinary regions. It adds no portal, via-region type, pathing fallback, or routing exception. Unrelated obstacles, unaligned overlaps, and single-layer regions keep the existing behavior.

## Reproduction and visualization

Stacked on #1874. Both PRs run the same fixture and write the same snapshot path; only the production code changes.

| Before | After |
| --- | --- |
| Before: the target/free area loses z5 | After: aligned ordinary regions keep the real layer stack |

Blue boxes are ordinary BGA capacity. Red boxes are same-net targets. A vertical blue line means one region is available on multiple layers; it is not a routed trace.

## Hosted validation

Nothing is run locally.

- focused fixture and updated SVG snapshot: pending GitHub Actions
- full repository test shards: pending GitHub Actions
- exact `srj24` sample 4 benchmark on the narrowed implementation: pending
- earlier broad split: failed after 432.3 s in selective rerip
- earlier uncovered-strip-only split: failed after 405.0 s in selective rerip

The PR stays draft until sample 4 solves, passes relaxed DRC, and its routed artifact has been visually checked.

### DIFF
--- a/lib/solvers/TopologyMergingSolver/get-cross-layer-target-access.ts
+++ b/lib/solvers/TopologyMergingSolver/get-cross-layer-target-access.ts
@@ -5,10 +5,14 @@ import {
   getCapacityMeshNodeBounds,
 } from "../TopologyPlanningSolver/capacity-node-geometry"
 import {
+  compactTopologyMergingRegions,
   doesBoundsContainPoint,
   getCanonicalCoordinates,
 } from "./topology-merging-regions"
-import type { TopologyMergingNodeGroup } from "./topology-merging-types"
+import type {
+  TopologyMergingNodeGroup,
+  TopologyMergingRegion,
+} from "./topology-merging-types"
 
 type CrossLayerAccessOverlap = {
   bounds: Bounds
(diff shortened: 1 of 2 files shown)


ShiboSoftwareDev:

ShiboSoftwareDev:

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 4000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 4000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 4000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 4000s --concurrency 1

ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --effort 1 --sample-timeout 4000s --concurrency 1

PR #1888, opened 2026-08-03
Preserve valid routes when differential-pair coupling is infeasible

seveibar:
## Summary
- wrap Pipeline7 differential-pair post-processing with an autorouter-level safe fallback
- preserve the already valid HD routes when the strict coupled router reports an infeasible pair
- reproduce the original core SOIC8 topology with pin1-to-pin1 and pin2-to-pin6, without changing fixture pins

## Why
Post-processing is an optimization after Pipeline7 has valid routed copper. An infeasible coupled-pair geometry should not fail the complete autoroute or encourage changing test pins to make the optimizer happy.

## Validation
- bun test tests/pipeline7-post-processing-asymmetric-terminal-fallback.test.ts
- bun test tests/pipeline7-post-processing-before-power-expansion.test.ts tests/pipeline7-post-processing-ambiguous-differential-pair.test.ts tests/pipeline7-post-processing-asymmetric-terminal-fallback.test.ts
- bun run build
- bun test --timeout 9999999: 441 pass, 55 skip, 0 fail
- core differential-pair routing-phase test with original pin selectors passes against the local build and regenerates two routed traces


seveibar:
CC @0hmX

PR #1889, opened 2026-08-03
Limit Pipeline7 power expansion to significant width requests

seveibar:
## Summary
- select power-expansion connections only when the requested nominal width is at least 0.1 mm wider and 2x the board minimum trace width
- preserve explicit onlyConnectionNames overrides for callers that intentionally want another trace expanded
- bypass signal-only boards byte-for-byte, preventing the power stage from rewriting differential pairs and ordinary signal traces
- pin the power-trace-expander dev dependency to the targeted-repair fix from tscircuit/power-trace-expander#13

## Snapshot impact
Seven existing signal-only Pipeline7 snapshots changed because the old power stage had been rewriting their route geometry. Only generated SVG snapshots were updated; no endpoints, pin selectors, or SRJ fixtures were changed.

## Validation
- bun run build
- bun test --timeout 9999999: 442 pass, 55 skip, 0 fail
- SRJ27 power expansion and DRC regression test passes
- signal-only e2e3 result is byte-for-byte identical before and after the power stage, with identical relaxed DRC error IDs

## Dependency
- tscircuit/power-trace-expander#13


seveibar:
/benchmark --dataset srj18 --sample-numbers 13 --concurrency 1

seveibar:
Benchmark conclusion for `482100c`: **not yet safe to merge**.

The [`/benchmark-all` sweep](https://github.com/tscircuit/tscircuit-autorouter/pull/1889#issuecomment-5170809710) found no relaxed-DRC regression on Default, SRJ18, SRJ19, SRJ21, or SRJ23. The SRJ18 sample 13 completion marker was runner contention: an [isolated rerun](https://github.com/tscircuit/tscircuit-autorouter/pull/1889#issuecomment-5170891631) solved in 335.6s versus main's 346.1s with the same 189 vias.

However, [SRJ20](https://github.com/tscircuit/tscircuit-autorouter/pull/1889#issuecomment-5170816639) has a real relaxed-DRC regression:

- completion: 73.0% → 73.5%
- relaxed DRC: 27.5% → 26.0%
- regressed samples: 61, 123, and 196

Diagnosis: those signal-only samples do not request significant width expansion. Main's power stage nevertheless ran its global via cleanup, relocating 6, 11, and 5 ordinary signal vias respectively and incidentally clearing their DRC errors. The new no-selection bypass correctly stops power expansion from touching those signals, but exposes missing generic via-clearance repair upstream.

Proposed fix: move/reuse the via-clearance repair as a generic post-routing stage, keep power expansion restricted to significantly wider traces, add focused SRJ20 regression tests, then rerun `/benchmark-all`.

PR #1900, opened 2026-08-03
test: reproduce sample 11 exact DRC residue

ShiboSoftwareDev:
## Issue

Dataset 18 sample 11 enters Pipeline 7’s production exact-DRC stage with 16 benchmark errors. The stage removes 14, but leaves one trace contact and one same-net via-clearance error, so the sample still fails DRC.

## Reproduction

The test runs the real sample through the production pipeline and snapshots the affected region before and after exact repair (`16 → 2`). This PR contains no solver or dependency change.

The fix is stacked in #1902.


ShiboSoftwareDev:
/benchmark-long --dataset 18 --sample-numbers 11 --concurrency 1 --sample-timeout 2000s

ShiboSoftwareDev:
/benchmark-long --dataset 18 --sample-numbers 11 --concurrency 1 --sample-timeout 2000s

PR #1917, opened 2026-08-04
Reduce Final-Route Vias with Transactional DRC-Safe Layer Collapsing

AnasSarkiz:
## Root cause

Pipeline 7's via simplification runs before width assignment and global/exact DRC repair, so final routes can retain safe same-layer excursions.

## Change

Adds `FinalViaOptimizationSolver` immediately after differential-pair post-processing and before power-trace expansion. It clones each eligible route, performs bounded deterministic same-layer excursion collapse, rebuilds transition metadata, converts through the production Pipeline 7 converter, then accepts only if production DRC, relaxed DRC, converted via count, trace length, route identity, terminals, and protected copper invariants all pass. Power expansion remains last and disallows new vias.

Differential pairs, power traces, preloaded traces, jumpers, through-obstacle routes, terminal-via routes, and via-in-pad routes are left unchanged.

## Validation

- `bun test` focused Pipeline 7 and final-optimization tests — pass
- `bun run build` — pass
- `bunx tsc --noEmit` is blocked by an existing `allowNewVias` assertion/type mismatch in `tests/pipeline7-post-processing-before-power-expansion.test.ts`.

## Benchmark status

Blacksmith benchmark execution was not available in this environment, so no benchmark before/after data is claimed. Required datasets: `dataset01`, `srj12`, `srj16`, `srj18`, `srj19`, `srj20`, `srj21`, `srj23`, `srj24`, and `srj27`.

| Dataset | Baseline completion / DRC / vias / time | Branch completion / DRC / vias / time |
| --- | --- | --- |
| Required suite | Not run here | Not run here |

Representative via reduction is covered by the focused clear `top → bottom → top` test.

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
@@ -797,8 +797,8 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
         }
         return [
           {
-            hdRoutes: cms.lengthMatchingPostProcessingSolver!.getOutput()
-              .hdRoutes,
+            hdRoutes:
+              cms.lengthMatchingPostProcessingSolver!.getOutput().hdRoutes,
             originalSrj: cms.originalSrj,
             obstacles: cms.srj.obstacles,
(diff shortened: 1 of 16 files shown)


AnasSarkiz:
Performance correction in c37827a: final via discovery now reuses one route spatial index, refreshes only after accepted mutations, and caps discovery at 512 high-transition routes. `/benchmark-all`

AnasSarkiz:
`/benchmark-long --dataset srj27`

PR #1921, opened 2026-08-04
Fix sample 10 mixed exact DRC residue

ShiboSoftwareDev:
## Issue

Dataset 18 sample 10 reaches the exact-DRC stage with a mixed cluster of trace and via-to-trace errors. The targeted clearance sweep selected only trace errors, so it could not repair the cluster in one accepted candidate.

## Implementation

Update high-density-repair03 so the existing targeted sweep also selects via-to-trace clearance errors. Existing force geometry and the strictly-better full-board DRC acceptance check are unchanged.

Repair PR: https://github.com/tscircuit/high-density-repair03/pull/52
Reproduction PR: https://github.com/tscircuit/tscircuit-autorouter/pull/1920

## Validation

Hosted sample-10 snapshot and benchmark pending.


ShiboSoftwareDev:
/benchmark-long --dataset 18 --sample-numbers 10 --concurrency 1 --sample-timeout 2000s

ShiboSoftwareDev:
Rejected by the hosted sample-10 benchmark: it still solves with 174 vias and the same 2 trace plus 2 via-to-trace errors. No follow-up patches will be stacked on this candidate.

PR #1925, opened 2026-08-04
Use the post-processing solver fallback

0hmX:
## Summary

- pin the post-processing solver revision that returns input HD routes when post-processing fails
- leave the autorouter pipeline and public output unchanged; it silently uses the returned fallback route
- update the existing bug regression to expect successful completion instead of an exception
- update the resulting visualization snapshot

The non-snapshot autorouter diff is exactly 10 changed lines: the dependency pin and three regression assertions.

## Validation

- bugreport82 regression — 1 pass
- package build, including declaration generation
- repository format check

Solver counterpart: tscircuit/length-matching-solver#39
Follow-up to: #1923


lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -87,6 +90,14 @@ interface CapacityMeshSolverOptions {
 }
 export type AutoroutingPipelineSolverOptions = CapacityMeshSolverOptions
 
+export type AutoroutingPipelineOutput = SimpleRouteJson & {
+  nonIdealRouteIssues?: NonIdealRouteIssue[]
+}
+
+export type GetAutoroutingPipelineOutputOptions = {
```
0hmX: I do not want anything to be newly needed to be added;

lib/autorouter-pipelines/AutoroutingPipeline7_MultiGraph/AutoroutingPipelineSolver7_MultiGraph.ts
```diff
@@ -1178,10 +1185,13 @@ export class AutoroutingPipelineSolver7_MultiGraph extends BaseSolver {
     })
   }
 
-  getOutputSimpleRouteJson(): SimpleRouteJson {
+  getOutputSimpleRouteJson(): AutoroutingPipelineOutput {
     return {
       ...this.originalSrj,
       traces: this.getOutputSimplifiedPcbTraces(),
+      autoroutingErrors:
```
0hmX: just ignore this place no need to so something special

PR #1928, opened 2026-08-05
Integrate fixed-topology portal-layer refinement

AnasSarkiz:
## Summary

- Assign every layer copy of a physical mesh-edge portal a stable `physicalPortGroupId` and serialize it into tiny-hypergraph.
- Run the fixed-topology portal-layer refiner after tiny-hypergraph section optimization in Pipelines 4, 7, 8, and 9.
- Explicitly lock preloaded synthetic trace assignments.
- Remove broad catch-based section-stage skip/fallback paths; the final tiny solution must come from a completed refinement stage. Completed no-op refinement preserves the original solved-solver identity, while accepted changes use the refined solver.
- Persist exact refiner counts, rejection reasons, predicted before/after demand, high-density via count, and per-phase timings through worker progress/results and benchmark JSON, including timeout snapshots.
- Render paired status classifications and paired solved metrics in GitHub benchmark comments; accepted-candidate rows are never hidden as unchanged.
- Pin `tiny-hypergraph` to `9cc3a682a8abf8ca6fd20144140716273f2f11ae` (companion draft: tscircuit/tiny-hypergraph#152).

## Current validation

- `bun test tests/benchmark-portal-refinement-stats.test.ts tests/features/pipeline9-minimal-preloaded-trace-graph.test.ts`: 2 passed, 0 failed, 77 expectations
- `bunx tsc --noEmit`
- Benchmark workflow YAML parses; all 3 embedded GitHub scripts pass `node --check`
- Real `dataset01/sample1` benchmark worker output includes the exact diagnostics in both the result and snapshot JSON
- `git diff --check`

## bugreport80 diagnosis

The failed CI rendering was not an accepted refiner change: `acceptedCandidateCount` is 0 and both outputs contain 76 vias. The CI-produced SVG is byte-for-byte the original `main` snapshot, while the prior expected snapshot was generated locally under Bun 1.3.9 and CI runs Bun 1.3.8. The branch now restores the original snapshot instead of blessing an unrelated rendering difference.

## Benchmark evidence and current gate

The previous `/benchmark-all` artifacts were mostly paired no-ops. The only apparent final-via change (`srj20/sample43`) also regressed relaxed DRC and is not reproducible locally: current branch and baseline both finish with 78 vias and zero relaxed-DRC errors while the refiner accepts zero candidates.

A targeted Blacksmith run for the real accepted-candidate case `srj18/sample6` is running on commit `00bcac3644b52db1b55cdcff754eb36691dfc83b`: https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31058429519

The expensive full `/benchmark-all` gate will be rerun only if targeted evidence shows a genuine final improvement without completion or DRC regression.


AnasSarkiz:
/benchmark --dataset srj18 --sample-numbers 6 --sample-timeout 600s --concurrency 1

AnasSarkiz:
/benchmark --dataset srj18 --sample-numbers 6 --sample-timeout 600s --concurrency 1

PR #1934, opened 2026-08-05
Validate safe trace via repair on affected samples

ShiboSoftwareDev:
## Issue

Dataset 18 sample 8 and Dataset 24 sample 2 are the current benchmark failures containing same-net via-clearance errors after exact DRC repair.

## Validation

Pin the focused repair from tscircuit/high-density-repair03#54 so both production samples can be rerun through Pipeline 7 in GitHub Actions.

The repair removes one option override: the safe trace branch now preserves the caller-enabled targeted sweep and can use the existing same-net via canonicalization.

Reproduction: tscircuit/high-density-repair03#53.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "zdwiel-dataset": "git+https://github.com/dwiel/tscircuit-benchmark.git#be36518b5bf51755dae92c230061ab3cf4e3e063",
     "high-density-repair01": "git+https://github.com/tscircuit/high-density-repair01.git#cefc7be547ff727bd31573ac096b850da847af46",
     "high-density-repair02": "https://codeload.github.com/tscircuit/high-density-repair02/tar.gz/2afc0cbba3bf2f7eb6b9cd33615d21e9ad9352d4",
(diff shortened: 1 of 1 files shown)


ShiboSoftwareDev:
/benchmark-long --dataset 18 --sample-numbers 8 --concurrency 1 --sample-timeout 2000s

ShiboSoftwareDev:
/benchmark-long --dataset 24 --sample-numbers 2 --concurrency 1 --sample-timeout 2000s

ShiboSoftwareDev:
## Exact result

The proposed repair has no effect on either production reproduction:

- Dataset 18 sample 8: 24 → 24 DRC errors; `pcb_via_clearance_error` 4 → 4; vias 301 → 301.
- Dataset 24 sample 2: 75 → 75 DRC errors; `pcb_via_clearance_error` 2 → 2; vias 331 → 331.

Both boards still solve but fail relaxed DRC. The Dataset 18 and 24 pass rates therefore remain 25% and 30%. This pin does not validate high-density-repair03#54 as the benchmark root fix.

PR #1932, opened 2026-08-05
Fix local layer access for srj24 sample 4

ShiboSoftwareDev:
## Problem

`srj24` sample 4 reaches a target pad on one layer while another topology continues underneath it on the other layer. The merged topology kept the neighboring free space split by layer, so the path graph could not leave the target area and change layers legally.

Two later invariants were also exposed by the same route:

- route simplification could keep a physical via but move the layer change to a different XY position;
- a portal could copy a same-net PCB terminal ID owned by another MST route.

```mermaid
flowchart LR
  A[Target on z0] --> B[Adjacent free rectangle]
  B --> C[Same rectangle is free on z1]
  C --> D[Legal local layer access]
  D --> E[Layer change stays at physical via]
  E --> F[Only this route's terminal IDs are emitted]
```

## Fix

- While generating atomic topology, join aligned free cells into one ordinary multi-layer region only when that rectangle borders a real target, continues the target topology on its layer, and continues a different topology underneath the target on the other layer.
- Keep the joined region only when the compacted rectangle is large enough for the configured via. All other aligned cells stay as their original per-layer regions before they can affect global compaction.
- When simplification changes layers, insert the same-XY transition at the physical via.
- Emit a portal's `pcb_port_id` only when that terminal is an endpoint of the current route.

This adds no solver stage, fallback, special region type, or post-processing repair. Unrelated board regions never enter the aligned representation.

## Same fixtures before and after

The reproduction PR #1931 and this PR run the exact same fixture sources. Only the production implementation changes.

### Local topology access

| Reproduction | Fix |
| --- | --- |
|  |  |

### Via transition reconstruction

| Reproduction | Fix |
| --- | --- |
|  |  |

### Terminal ownership

| Reproduction | Fix |
| --- | --- |
|  |  |

## Hosted validation

- The reproduction stack's full Bun test run is green: [run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31053277164).
- On this exact topology SHA, `srj24` sample 4 passes static reachability and reaches Tiny graph search. It then exposes the separate fixed-budget issue after 157.9s: [run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31057579418).
- The previously regressed `srj18` sample 6 now solves with 307 vias in 890.8s instead of timing out in trace simplification: [run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31057578589).
[...]

PR #1935, opened 2026-08-05
[diagnostic] isolate sample 4 via transition behavior

ShiboSoftwareDev:
Temporary hosted-only diagnostic for #1932. This keeps the topology and terminal-ownership fixes but removes the via-transition reconstruction, so the existing snapshot suite can identify which change caused the dataset 18 sample 11 exact-DRC regression. This PR will be closed after the result is recorded.


ShiboSoftwareDev:
Isolation complete: sample 11 remains 7 → 5 without the via-transition reconstruction, so the regression is caused by the topology alignment change. Closing this temporary diagnostic.

PR #1939, opened 2026-08-06
Scale Tiny search budget with graph size

ShiboSoftwareDev:
## Problem

#1938 reproduces sample 4 reaching Tiny graph pathing's normal 2M limit without a complete result. Increasing every large board's initial limit changed boards that already had an acceptable 2M result, including `bugreport68`.

## Change

Keep the initial search and acceptance behavior unchanged. Only continue when the existing best-result and greedy acceptance both fail:

```mermaid
flowchart LR
  A[Run normal 2M search] --> B[Existing final acceptance]
  B -->|complete result| C[Return unchanged]
  B -->|still unresolved| D[Extend the same search once]
  D --> E[route count × 32k × effort]
```

The continuation uses the solver's current in-memory state; it does not restart routing or add a pipeline stage. The existing wrappers reserve the same upper bound so they cannot stop the active child early. For sample 4's 841 routes, the final cap becomes 26.912M iterations at effort 1.

The fix PR changes only the continuation calculation. It does not alter topology, costs, route geometry, solver selection, or fallback behavior. There is no graph-size threshold: already-accepted boards stop exactly where they did before, while only an unresolved search receives more work.

## Visual fixture

This PR reuses #1938's exact test and fixture unchanged. The ordinary graph remains at 2M because it has an accepted result. Sample 4's unresolved search continues to 26.912M, so the previously failing assertion passes.

| Reproduction | This PR |
| --- | --- |
| The unresolved sample-sized search remains limited to 2M iterations | The same unresolved search receives a 26.912M continuation allowance |

## Hosted result

- All five test shards and focused SVG verification pass on the final commit: [test run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31069465998).
- `srj18` sample 3 solves in 43.8s and sample 6 solves in 916.1s: [benchmark](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31062677107).
- `srj24` sample 2 solves in 652.3s and sample 3 solves in 1913.5s: [benchmark](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31062679033).
- `srj24` sample 4 keeps progressing past 2M, but still times out at 4000s after about 18.7M phase iterations in that same benchmark.

This removes the fixed-budget limitation, but it is **not sufficient by itself to make sample 4 pass**. The remaining issue is the cost of searching the final routes, so this PR remains a draft while that root performance problem is investigated.

No solver, snapshot, or benchmark tests were run locally.


ShiboSoftwareDev:
/benchmark-long --solver AutoroutingPipelineSolver7_MultiGraph --dataset srj18 --sample-numbers 3,6 --effort 1 --sample-timeout 2000s --concurrency 2

ShiboSoftwareDev:
/benchmark-long --solver AutoroutingPipelineSolver7_MultiGraph --dataset srj24 --sample-numbers 2,3,4 --effort 1 --sample-timeout 4000s --concurrency 3

ShiboSoftwareDev:
/benchmark-long --solver AutoroutingPipelineSolver7_MultiGraph --dataset srj18 --sample-numbers 3,6 --effort 1 --sample-timeout 2000s --concurrency 2

ShiboSoftwareDev:
/benchmark-long --solver AutoroutingPipelineSolver7_MultiGraph --dataset srj24 --sample-numbers 2,3,4 --effort 1 --sample-timeout 4000s --concurrency 3

PR #1946, opened 2026-08-06
fix(topology): drop sub-epsilon obstacles that abort topology merging

### LINKED ISSUE #1614: TopologyMergingSolver crashes on zero-area "global" mesh node ("invalid bounds"), emptying the PCB preview
### Summary

`TopologyMergingSolver` throws `node "cmn_0" in group "global" has invalid bounds` when a `global` capacity-mesh node produced by the upstream topology-planning stage has degenerate (zero/near-zero area) bounds. In production `runframe` sessions this aborts the whole autorouting pipeline and leaves the PCB preview empty, and the same error is re-thrown on every render attempt (observed firing dozens of times in a single session).

### Where it throws

`lib/solvers/TopologyMergingSolver/topology-merging-input.ts` (`validateTopologyMergingNode`):

```
TopologyMergingSolver: node "cmn_0" in group "global" has invalid bounds
```

`isValidCapacityBounds` rejects the node because its width or height is `<= GEOMETRY_EPSILON` (or non-finite). The offending node is part of `topologyOutput.globalMeshNodes`, so the degenerate geometry originates upstream in the global mesh / topology-planning stage rather than in the merging solver itself.

Observed with `@tscircuit/capacity-autorouter@0.0.665`; the throw is unchanged on `main`.

### Why not just swallow it

The repo's `AGENTS.md` "Fallback Logic" policy is explicit that a solver should fix the root cause or throw, not silently drop bad input and continue (that yields a "plausible-looking wrong route"). So the fix should be in whatever produces a zero-area `global` node, not a filter in the merging solver. Filing this so the root cause can be found with a proper reproduction.

### What's needed

- A minimal circuit / `SimpleRouteJson` that reproduces a zero-area `global` mesh node reaching `TopologyMergingSolver`.
- Fix the upstream stage so it never emits a degenerate global node (or, if a zero-area node is ever legitimate, decide the correct policy explicitly).

### Related

The fact that this bubbled up as a repeated JS exception on the `runframe` side (rather than being shown as an autorouting failure) is being addressed separately in `runframe`.

zkasuran:
Fixes #1614

## Repro

Two tests, both failing on `main` before the fix.

`tests/features/topology/merged-topology/degenerate-global-obstacle-nodes.test.ts`
builds a small two layer board whose obstacle list contains one rect `5e-10` wide
and one `1e-10` tall. Topology planning turns them into zero-area capacity nodes
and the pipeline aborts with the error from the issue:

```
TopologyMergingSolver: node "cmn_12" in group "global" has invalid bounds
```

`tests/features/topology/merged-topology/degenerate-component-obstacle-nodes.test.ts`
does the same for a BGA whose member pads include one sub-epsilon pad. That one
fails in the component-local group:

```
TopologyMergingSolver: node "free-u_bga-u_bga_dust-1" in group "component-0" has invalid bounds
```

## Root cause

Two stages disagree about what counts as geometry.

- RectDiff drops a rect only when a dimension is non-positive (`w <= 0 || h <= 0`).
- `isValidCapacityBounds` requires both dimensions to be strictly greater than
  `GEOMETRY_EPSILON`, which is `1e-9`.

An obstacle whose width or height lands in `(0, 1e-9]` passes the first gate and
fails the second. It survives into `topologyOutput.globalMeshNodes` (or into a
component group through the BGA/QFP/SOIC generators) and `TopologyMergingSolver`
input validation throws on it. An obstacle dimension of exactly `0` is already
dropped upstream, so the window is narrow and easy to miss.

Two details the repro corrects in the report: the degenerate node does not
originate inside the global mesh stage, it is the obstacle rect handed to that
stage (the failing node carries exactly the obstacle's dimensions). The same
input also produces degenerate component-local nodes, which the report does not
mention.

## Fix

Topology planning drops obstacles that are degenerate under the same predicate
the capacity node validation uses, at its input, before any topology is
generated. `dropDegenerateObstacles` sits next to `isValidCapacityBounds` so the
two definitions cannot drift. `normalizeInput` applies it to the SRJ that every
downstream stage derives from, which covers the global RectDiff solve and the
component-local solves.

This is an explicit policy, not a swallowed error. A rect thinner than a
picometre carries no routable and no blocking area: it cannot hold a port point
and no trace can collide with it, so removing it cannot change a route. Nothing
has failed at normalization time. The loud validation in `TopologyMergingSolver`
is untouched, so it still throws for every other invalid input.

#1615 went the other way and filtered nodes after the generators had produced
them. It was closed by the stale bot without review. Removing the geometry at the
input is what the issue asked for and it covers the component groups as well.

## Behaviour on existing boards

No board in the repo carries such an obstacle, so the filter is a no-op
everywhere except on the broken input in the issue. Measured by scanning every
SRJ the repo can route:

| corpus | SRJs | obstacles | in `(0, 1e-9]` |
| --- | --- | --- | --- |
| `fixtures/**/*.json` | 189 | 39,039 | 0 |
| bundled benchmark datasets | 1,012 | 136,654 | 0 |

## Verified locally

- `bun test tests/features/topology tests/component-detection-bga-only.test.ts`:
  35 pass, 0 fail, 21 files
- the two new tests: 2 pass with the fix, both fail on an unmodified checkout of
  the parent commit with the node ids quoted above
- `bun test tests/pipeline7-final-net-colors.test.ts
  tests/pipeline7-lock-terminal-ports.test.ts
  tests/features/arduino-uno-inner-pours.test.ts
  tests/srj-obstacle-component-id.test.ts tests/e2e1.test.ts`: 6 pass, 1 skip,
  0 fail
- `bunx tsc --noEmit`: clean
- `bunx biome format` on the five touched files: clean
- `bun run build`: success

Happy to split this into a repro PR and a fix PR if you prefer that order; the
two commits are already separated that way.

AI assistance (Claude, Anthropic) was used in developing this change. The design,
review and verification were done by the author.

### DIFF
--- a/lib/solvers/TopologyPlanningSolver/MultiGraphTopologyPlannerSolver.ts
+++ b/lib/solvers/TopologyPlanningSolver/MultiGraphTopologyPlannerSolver.ts
@@ -187,7 +187,7 @@ export class MultiGraphTopologyPlannerSolver extends BasePipelineSolver<MultiGra
   private getComponentNoConnectionSrjs(): SimpleRouteJson[] {
     return this.normalizedInput.components.map((component) =>
       createComponentSrj({
-        inputSrj: this.inputProblem.inputSrj,
+        inputSrj: this.normalizedInput.inputSrj,
         component,
       }),
     )
--- a/lib/solvers/TopologyPlanningSolver/capacity-node-geometry.ts
+++ b/lib/solvers/TopologyPlanningSolver/capacity-node-geometry.ts
@@ -40,6 +40,28 @@ export function isValidCapacityBounds(bounds: Bounds): boolean {
   return hasValidWidth && hasValidHeight
 }
 
+/**
+ * Removes obstacles that are thinner than GEOMETRY_EPSILON in either dimension.
+ *
+ * Topology generation turns every obstacle rect into capacity mesh nodes, and a
+ * rect below GEOMETRY_EPSILON produces a node that `isValidCapacityBounds`
+ * rejects, so it aborts TopologyMergingSolver input validation. Such a rect also
+ * carries no routable or blocking area: it cannot hold a port point and no trace
+ * can collide with it. Topology planning drops it at its input so every stage
+ * uses one definition of real geometry.
+ */
+export function dropDegenerateObstacles(obstacles: Obstacle[]): Obstacle[] {
+  return obstacles.filter((obstacle) => {
+    const obstacleBounds = getBoundFromCenteredRect({
+      center: obstacle.center,
+      width: obstacle.width,
+      height: obstacle.height,
+    })
+
+    return isValidCapacityBounds(obstacleBounds)
+  })
+}
+
 export function isNodeInsideOrOverlappingObstacle({
   node,
   obstacle,
--- a/lib/solvers/TopologyPlanningSolver/topologyPlanningShared.ts
+++ b/lib/solvers/TopologyPlanningSolver/topologyPlanningShared.ts
@@ -24,12 +24,14 @@ import "lib/solvers/BgaTopologyGeneratorSolver/BgaTopologyGeneratorSolver"
 import "lib/solvers/QfpThermalPadTopologyGeneratorSolver/QfpThermalPadTopologyGeneratorSolver"
 import "lib/solvers/QfpTopologyGeneratorSolver/QfpTopologyGeneratorSolver"
 import "lib/solvers/SoicTopologyGeneratorSolver/SoicTopologyGeneratorSolver"
+import { dropDegenerateObstacles } from "./capacity-node-geometry"
 import type {
   MultiGraphTopologyPlannerSolverParams,
   SerializedTopologyComponentInput,
 } from "./MultiGraphTopologyPlannerSolver"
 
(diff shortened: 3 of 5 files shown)


zkasuran:
A nudge before this goes quiet, since it has had no review yet. All checks are green: build, format-check, type-check and all 5 test shards pass.

The change only drops sub-epsilon (zero-area) obstacles at the input to topology planning, using the same `isValidCapacityBounds` predicate the node validation already applies, so it targets exactly the degenerate obstacles that abort the pipeline today. I re-checked the no-op scope: 0 of 39,039 fixture obstacles and 0 of 136,654 dataset obstacles fall in the degenerate window it touches, so there is no benchmark risk. Two repro tests cover the global group and the component group.

Happy to rebase or adjust if you want any change.

zkasuran:
Re-verifying before this goes quiet, no review yet. Head 5602508c, CI green (12 pass, 2 optional profile jobs skipped), mergeStateStatus CLEAN, no conflicts. This only drops sub-epsilon obstacles that abort topology merging. It is a proven no-op on real geometry: zero of 39,039 fixture obstacles and zero of 136,654 benchmark obstacles fall in the sub-epsilon window it touches. Ready for review or a /benchmark run if you want fresh numbers.

PR #1951, opened 2026-08-06
Fix conservative approximation of rotated pads

imrishabh18:
## Issue

Previously it used to touch pad of different net

image

image

## Summary

- replace the sparse rotated-pad centerline approximation with conservative clipped world-axis strips
- keep the real rotated pad area blocked without expanding to a full 2D approximation grid
- add sparse-geometry and strict DRC regressions for `bugreport66` / `pcb_port_115`
- refresh the four affected routing snapshots

## Root cause

When a board had more than 20 connected obstacles, the sparse approximation represented an elongated rotated pad with centerline squares based on its shorter dimension. That omitted portions of the rotated corners, so the router treated real copper around `pcb_port_115` as free space and routed a different net through it. Strict DRC then correctly reported the overlap.

## Validation

- focused rotated-pad geometry and `bugreport66` regression tests
- affected snapshot tests for `bugreport71`, `bugreport80`, and `pipeline9`
- every failure reported by the initial full-suite run was revalidated with adequate permissions/timeouts; no remaining known regressions


seveibar:
This is why fallbacks are bad, we should always be using the same function to approximate obstacles, we have a conditionally invoked global approximation that is being fixed here but we should remove the condition

seveibar (approved):
See comment , ideally remove the conditional call

PR #1954, opened 2026-08-06
Inject explicit initial connectivity into point pairing

seveibar:
## What changed

- builds the canonical `initiallyConnectedMap` once at each pipeline boundary from the exact SRJ passed to point pairing
- requires that map in both net-to-point-pair solvers
- adapts the map's equivalence classes into zero-cost MST edges
- removes local connectivity reconstruction from `externallyConnectedPointIds` and `srj.traces[].connectsTo`
- avoids geometric connectivity inference and prevents the solver from owning a second, partial connectivity model

## Validation

- `bun test tests/solvers/net-to-point-pairs-explicit-initial-connectivity.test.ts tests/solvers/net-to-point-pairs-root.test.ts tests/utils/get-initially-connected-map-from-simple-route-json.test.ts`
- `bunx tsc --noEmit`
- `bun run format:check`
- `bun run build`
- top of stack: `bun test tests/features/pipeline9-srj23-solved-visual.test.ts`

## Stack

1. **This PR: canonical initial connectivity boundary and MST adapter**
2. Immutable via anchors
3. Pipeline9 B01 routing and regional repair layers


seveibar:
Incorporated the connectivity-layer feedback: the SRJ fallback has been removed, both point-pair solvers now require the canonical upstream initiallyConnectedMap, and their only responsibility is adapting its equivalence classes for the MST. Every pipeline constructs the map from the exact SRJ passed into point pairing. The full dependent PR stack was rebased and force-pushed after this change.

seveibar:
CI fix: the canonical transitive connectivity map intentionally changes the Pipeline9 SRJ23 and Pipeline7 SRJ27 routes, so I updated both Linux/non-Linux visual baselines and the explicit SRJ27 DRC expectations. All six SRJ27 samples now pass relaxed DRC both before and after power expansion (previously 1/6 before and 4/6 after). Fresh run 31136681421 passed all five Bun shards; build, type-check, format-check, Vercel build, and deployment also pass. The full stacked PR chain has been rebased and repushed.

seveibar:
Fixed the blank Pipeline7 power-trace-expansion snapshots. The canonical initial connectivity made these fully routed SRJ27 inputs emit no new routes, so the power stage previously received no mutable traces and the visual rendered an empty routed-only result. Selected preloaded power traces are now passed into expansion as explicit replacements, while unselected traces remain fixed. The visual renders the joint pre/post board state and asserts both panes are nonempty. All six SRJ27 visual samples retain zero relaxed-DRC errors, and all CI checks now pass.

seveibar:
Rebased onto current main and resolved the Pipeline7 sample003 snapshot conflict against the combined behavior. The focused SRJ27 visual test still passes all six samples with nonempty routed pre/post panes. I also replayed and force-updated the full dependent stack; GitHub reports every PR mergeable. #1954 is CLEAN and all checks pass.

PR #1955, opened 2026-08-06
Support immutable via anchors during trace simplification

seveibar:
## What changed

- includes vias from immutable `otherHdRoutes` as same-net merge anchors
- never mutates or re-emits immutable anchor vias
- accepts explicit net metadata for synthetic connection names
- passes that metadata through `TraceSimplificationSolver`

## Validation

- `bun test --timeout 9999999 tests/solvers/same-net-via-merger-immutable-anchor.test.ts`
- `bun test --timeout 9999999 tests/solvers/same-net-via-merger-direct-overlap-star.test.ts tests/features/same-net-via-clearance-repair-visual.test.ts`
- `bunx tsc --noEmit`

## Stack

- Depends on #1954
- Next: Pipeline9 B01 routing

### DIFF
--- a/lib/solvers/SameNetViaMergerSolver/SameNetViaMergerSolver.ts
+++ b/lib/solvers/SameNetViaMergerSolver/SameNetViaMergerSolver.ts
@@ -16,6 +16,8 @@ export interface SameNetViaMergerSolverInput {
   inputHdRoutes: HighDensityRoute[]
   /** Routed copper that participates in collision checks but is never changed. */
   otherHdRoutes?: ReadonlyArray<HighDensityRoute>
+  /** Explicit connection metadata for routes whose names are not in connMap. */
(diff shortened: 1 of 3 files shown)


lib/solvers/SameNetViaMergerSolver/SameNetViaMergerSolver.ts
```diff
@@ -338,6 +381,7 @@ export class SameNetViaMergerSolver extends BaseSolver {
                   mergedViaHdRoutes: this.mergedViaHdRoutes,
                   hdRouteSHI: this.hdRouteSHI,
                   obstacleSHI: this.obstacleSHI,
+                  netByConnectionName: this.netByConnectionName,
```
seveibar, line 384: i don't think you should need netByConnectionName, isn't the connMap good enough?

PR #1956, opened 2026-08-06
Route Pipeline9 regions around immutable copper with B01

seveibar:
## What changed

- replaces Pipeline9's regular high-density stage with a small B01-only adapter
- loads preloaded routes as layer-aware immutable B01 route obstacles
- explicitly injects serialized initial connectivity into Pipeline9 point pairing
- rejects routing nodes larger than 15x15 mm
- fails clearly when B01 cannot solve; this layer contains no reroute fallback
- adds `@tscircuit/high-density-b01` as a development dependency

## Validation

- `bun test --timeout 9999999 tests/features/pipeline9-minimal-preloaded-trace-graph.test.ts tests/features/pipeline9-srj23-sample58-fixed-copper-clearance.test.ts`
- `bunx tsc --noEmit`
- `bun run build`

## Stack

- Depends on #1955
- Next: regional reroute/splicing fallback


seveibar:
Updated Pipeline9 to keep the existing force-improve and high-density repair stages unchanged after B01. B01's fixed-copper contract is still tested at the B01 boundary; any interaction introduced by the legacy postprocessors is handled by the downstream joint-DRC repair stack. I also updated this layer's visual fixtures to samples B01 itself solves before the fallback PR is applied. All #1956 checks are green.

PR #1984, opened 2026-08-07
ci: balance test shards by recorded runtime, scale to 8 machines

seveibar:
## Problem

`Bun Test` is the only slow check in this repo — median **10.2 min**, p90 **14.9 min**. Everything else (Type Check, Build, Format, Vercel) finishes in under 2 minutes.

The 5-way matrix is imbalanced, and always in the same direction. Per-shard means over 16 recent runs:

| shard | mean | median | max |
|---|---|---|---|
| test (1) | 7.3m | 7.2m | 9.0m |
| test (2) | 8.9m | 8.5m | 14.9m |
| **test (3)** | **10.4m** | **11.0m** | **13.0m** |
| test (4) | 6.1m | 5.8m | 8.3m |
| test (5) | 4.5m | 4.2m | 7.9m |

Average total work is 37.2 machine-minutes, so a perfectly balanced 5-way split would finish in **7.4 min** rather than 11.0. Roughly a third of current wall time is pure imbalance.

Setup (checkout, bun, `bun install`, `npm install -g`) is only ~25s per shard, so dependency caching isn't worth pursuing — it can't pay back more than about 20 seconds.

## Root cause

`@tscircuit/bun-test-plan` distributes files **round-robin over an alphabetical sort**, with no runtime weighting (`nodeIdx = idx % nodeCount`). Cost in this repo is extremely skewed:

```
186.6s  tests/bugs/bugreport46-ac4337-no-near-gap-multiroutes.test.ts
183.0s  tests/bugs/bugreport78-dfeaaf.test.ts
167.0s  tests/features/pipeline7-power-trace-expansion-visual.test.ts
128.1s  tests/bugs/bugreport77-07f6a7.test.ts
124.0s  tests/bugs/bugreport68-41562e.test.ts
```

**10 of 354 files are 55% of the 36.9 min total.** The heavy `bugreport*` files are alphabetically adjacent, so the modulus drops several of them on the same shard every single time. `bun-test-plan.json` already pins `bugreport23` into its own glob pattern — that hand-patched exactly one file's worth of this problem.

A round-robin simulation reproduces the observed result to within 1% (predicts 12.14 min for the run measured at 12.0), which is what validates the cost model below.

## Why more machines alone doesn't fix it

The imbalance is a resonance between file order and the modulus, so adding machines is erratic and **not monotonic**:

| machines | round-robin (today) | LPT balanced |
|---|---|---|
| 5 | 12.14m | **7.73m** |
| 6 | 12.34m ← *slower than 5* | 6.53m |
| 8 | 7.28m | **5.02m** |
| 10 | 8.63m ← *slower than 8* | 4.12m |
| 12 | 7.96m | 3.61m |
| 16 | 4.90m | 3.61m |

Going 5 → 6 machines would make CI **slower**. Balancing is the change that matters; extra machines only pay off once it's in place.

## What this PR does

1. **Replaces round-robin with LPT bin packing** (`scripts/generate-balanced-test-plan.ts`) — sorts files by recorded cost descending, assigns each to the least-loaded shard. Achieves a **0.00 min spread** on real data. Also drops a global `npm install` from every shard.
2. **Checks in `test-timings.json`** — measured per-file wall time, parsed from CI logs by `scripts/record-test-timings.ts`. Timings are stable run to run (median ratio 1.00, p90 1.05 across two runs), so this rarely needs regenerating.
3. **Takes the matrix from 5 to 8 machines.** The repo is public so runners are free, and jobs aren't queueing today (median queue delay 3s). Costs ~3 extra machine-minutes per run.
4. **Adds a `concurrency` group with `cancel-in-progress` for PRs.** `bun-test.yml` had none, and three recent PRs each ran 2-3 full Bun Test runs because pushes superseded each other. This saves machine time rather than wall time. `main` runs are not cancelled.

**Net effect: ~12.0 min → ~4.7 min critical path**, a ~60% cut. At the current ~67 Bun Test runs/day that's roughly 7 hours/day of reclaimed waiting.

## Where the floor is

**Don't scale past ~10-12 machines.** `bugreport46` is a single atomic `test()` call taking 187s, so no sharding can beat **~3.6 min** (187s + 30s setup). At 12 machines that one file is alone on its shard; beyond 12, added machines do nothing at all.

Breaking below that requires making the tests themselves faster, which is a code change, not a CI change.

## Known limitation

New test files have no recorded timing and are costed at the p90 of known files (~12s). Because the distribution is so skewed, **no default fully protects balance** — a new file that genuinely takes 187s will push its shard from 4.7 to ~7.5 min regardless of what we guess. The default bounds the damage (still better than the 12 min this replaces) and stops a batch of new files from clustering; refreshing `test-timings.json` is the real fix. The generator prints every unmeasured file it encounters, so this is visible in the log rather than silent.

## Validation

- Cost model reproduces observed shard times to within 1%
- Packer output on current `main`: 8 shards, 4.69 min each, **spread 0.00 min**
- Shard coverage is exact: 357 files, no duplicates, none missing
- Both scripts pass the repo's own `bunx tsc --noEmit`
- `bunx biome format .` clean repo-wide (only the two new files were formatted)
- Ran files from a generated plan end to end: 3 pass, 0 fail

## Follow-ups (not in this PR)

- `bugreport68-41562e` has an internal 200s timeout but takes 124-260s, so it flakes on timeout — it caused the failures in runs `31178537084` and `31181521101`.
- The LPT packer could go upstream into `@tscircuit/bun-test-plan` as optional timing-weighted distribution, so other tscircuit repos benefit. Kept local here to keep this PR self-contained and reversible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


seveibar:
## Measured on this PR

This PR changes the workflow that runs on it, so the first CI run is the experiment. [Run 31204498688](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31204498688) — all 8 shards green:

| shard | duration |
|---|---|
| test (1) | 4.82m |
| test (2) | 4.53m |
| test (3) | 3.35m |
| test (4) | 4.77m |
| **test (5)** | **5.13m** |
| test (6) | 4.60m |
| test (7) | 4.62m |
| test (8) | 4.35m |

|  | before | after |
|---|---|---|
| critical path | 12.0m | **5.13m** |
| shard spread | 8.2m | **1.78m** |
| machine-minutes | 37.2m | 36.2m |

**57% faster**, and machine-minutes did not increase — the extra 3 machines are paid for by the shards no longer idling.

The prediction from the offline cost model was 4.69m packing + ~0.5m setup = 5.19m; actual was 5.13m, so the model was accurate to within ~1%.

Residual spread is 1.78m rather than the 0.00m the packer targets, from two sources: runner-to-runner speed variance, and the three test files added to `main` since the timings were recorded (estimated at p90 rather than measured). Both shrink on the next `test-timings.json` refresh. Nothing here needs the shard count changed.

All 17 checks pass.

PR #2008, opened 2026-08-08
Stop coalesced vias from changing anchors

ShiboSoftwareDev:
## Problem

Via merging can move a via onto an anchor that already contains another same-net via. On the next iteration it may move one member of that shared anchor toward a different nearby anchor, creating a two-state loop.

## Fix

Count same-net vias at each coordinate before proposing moves. A via at a coordinate already shared by another same-net via is already physically coalesced, so it is not moved independently again.

This does not add a fallback or force a guessed position. Unshared vias still use the existing obstacle-aware merge choice.

## Visual fixture

The orange via joins the left shared anchor and remains there

Purple circles are two existing shared anchors. The orange via is within merge range of both. The result shows it coalesced at one anchor instead of alternating forever.

## Validation

- `bun test --timeout 9999999 tests/solvers/same-net-via-merger-coalesced-anchor.test.ts`
- final stack: `bun run build`

The full `srj24` sample 4 benchmark is running on this final stacked PR with one sample and one worker: [run 31240277219](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/31240277219).


ShiboSoftwareDev:
/benchmark-long --pipeline 7 --dataset 24 --sample-numbers 4 --concurrency 1 --sample-timeout 7200s

PR #2026, opened 2026-08-08
Add autorouting bug report 76a6de regression

imrishabh18:
## What changed

- Add the captured autorouting bug report `76a6de9f-6255-4cf2-ab9d-b8b0201033fc` as `bugreport87-76a6de`.
- Add the generated debugger fixture and SVG regression snapshot.
- Allow the focused test 120 seconds because this 96-trace capture takes about 96 seconds to solve.

## Why

This preserves the reported board as a deterministic regression case for future autorouter changes. The current solver reproduces the captured route successfully; the longer timeout prevents Bun's default 5-second timeout from turning the case into a false CI failure.

## Validation

- `bun run format:check`
- `bun test tests/bugs/bugreport87-76a6de.test.ts`


tests/bugs/bugreport87-76a6de.test.ts
```diff
@@ -0,0 +1,21 @@
...
+test(
+  "bugreport87-76a6de.json",
+  () => {
+    const solver = new AutoroutingPipelineSolver(srj)
+    solver.solve()
+    expect(getLastStepSvg(solver.visualize())).toMatchSvgSnapshot(
+      import.meta.path,
+    )
+  },
+  { timeout: 120_000 },
```
imrishabh18:
```suggestion
  { timeout: 180_000 },
```

PR #2038, opened 2026-08-09
Add via arrays for wide traces

imrishabh18:
## Before

image

## After

image

## Summary

- install `@tscircuit/power-trace-expander` directly from [tscircuit/power-trace-expander@05e8d89](https://github.com/tscircuit/power-trace-expander/commit/05e8d89ca8326664fd4e8c0f79060aab70444ee0) ([upstream PR #17](https://github.com/tscircuit/power-trace-expander/pull/17))
- enable the upstream `addViaArrays` finalization pass by default in Pipeline 7 while preserving an explicit caller override
- remove the duplicated autorouter-side via-array implementation
- retain bugreport87 structural coverage and add a before/after relaxed-DRC regression guard
- update affected Pipeline 7 snapshots

## Root cause

Pipeline 7 expanded wide power copper while retaining a single minimum-size transition via. The initial autorouter implementation added rows after power expansion, outside the expander's obstacle and clearance context.

The upstream finalizer sizes rows from the shared adjacent trace width, checks both-layer copper containment and all-board obstacles, preserves via spacing, rotates blocked rows locally, and keeps the original via when no complete safe row fits.

## Impact

Wide power-trace layer transitions can receive multiple clearance-safe vias. `addViaArrays` is independent of `allowNewVias`; Pipeline 7 enables the former while continuing to disable arbitrary new routing vias.

## Validation

- `bun run format:check`
- `bunx tsc --noEmit --pretty false`
- `bun run build`
- bugreport87: adjacent via array present and post-expansion relaxed DRC does not exceed pre-expansion DRC
- SRJ27 samples 001–006: all remain at zero relaxed DRC; sample001 confirms an array was added
- full local suite: 493 passed, 57 skipped; three feature-affected snapshots were refreshed
- one unrelated macOS exact-DRC visualization differs identically on untouched `main`; its exact DRC output remains zero


seveibar (approved):
Im pretty sure this is going to make traces discontinuous and potentially create issues with tscircuit/checks- which is fine but it will need to be fixed downstream. Trace endpoints will not be continuous with this implementation.

im not sure i would implement the trace array like this- which guideline did you follow?

PR #2043, opened 2026-08-09
fix(pipeline7): propagate adaptive high-density budgets

ShiboSoftwareDev:
Depends on:

- tscircuit/high-density-a01#87 — exact `srj24` sample 1 reproduction
- tscircuit/high-density-a01#88 — direct A01 fixes

## Problem

The dense region in `srj24` sample 1 is still adding completed routes when A01 reaches its initial estimated budget. The A01 fix extends its own deadline while that progress continues, but the surrounding autorouter portfolio had already copied the old deadline and could stop first.

## Fix

- consume the fixed A01 commit
- when the candidate that just ran changes its iteration limit, recompute the portfolio limit from the candidates' remaining work

The portfolio does not receive a larger fixed budget and does not poll every candidate on every step. It reacts only when the active child changes its own limit.

## Visual reproduction

The exact same fixture and layer-by-layer snapshot is used on both sides of the A01 stack:

| Before | After |
| --- | --- |
| A01 stops after 31 of 46 routes | A01 completes all 46 routes |

Each panel is the same physical region on one PCB layer. Colored dots are terminals, colored paths are traces on that layer, and gray circles are through-vias.

## Verification

- focused portfolio budget-propagation regression test passes
- autorouter production build passes
- A01 package suite: 59 passed, 2 skipped, 0 failed
- exact dense-region fixture: 46/46 routes, 0 geometry violations

The end-to-end Pipeline 7 result will be recorded below by the GitHub `srj24` sample 1 benchmark; no local full-dataset benchmark was used.

### DIFF
--- a/lib/solvers/HyperHighDensitySolver/PortfolioSingleIntraNodeSolver.ts
+++ b/lib/solvers/HyperHighDensitySolver/PortfolioSingleIntraNodeSolver.ts
@@ -365,10 +365,20 @@ export class PortfolioSingleIntraNodeSolver extends HyperParameterSupervisorSolv
       this.expandAdaptiveSearch()
     }
 
+    const activeCandidate = this.getSupervisedSolverWithBestFitness()?.solver
+    const activeCandidateIterationLimit = activeCandidate?.MAX_ITERATIONS
     super._step()
 
     if (!this.solved && !this.failed && this.shouldExpandPortfolio()) {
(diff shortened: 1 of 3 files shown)


ShiboSoftwareDev:
/benchmark-long --dataset 24 --sample-numbers 1 --sample-timeout 2000s --concurrency 1

PR #2084, opened 2026-08-12
Fix nested cramped escapes for bugreport91

imrishabh18:
## Summary

- search one level deeper when a sparse, unmerged source terminal has no non-cramped escape
- retain at least one candidate from each distinct local cramped branch and prefer exits into higher-capacity regions
- preserve legacy selection for dense/internal nodes, merged source-net regions, and preloaded traces
- replace bugreport91's terminal failure image with a final, fully routed board snapshot
- assert end-to-end pipeline completion and verify every generated point-pair has a final high-density route
- add focused coverage for branch preservation and merged-terminal legacy behavior

## Root cause

Bugreport91 has six adjacent connector routes whose legal exits cross two consecutive cramped boundaries. The necessary-cramped-port solver searched only two port hops and then kept the first five candidates globally. That could remove the second boundary or an entire equal-cost local branch, making those routes structurally unreachable. Global reripping could not recover an edge that had already been removed from the graph.

## Impact

Bugreport91 now completes the entire autorouting pipeline with every generated point-pair represented in the final high-density routes. The deeper branch-preserving behavior is limited to sparse, unmerged source terminals so existing complex and preloaded topologies retain their previous graph.

## Validation

- paired SRJ18 same-machine benchmark: 0 outcome regressions, completion 56.3% → 56.3%, timeouts 7 → 7, P50 263.3s → 261.1s, average vias unchanged at 166.22
- `bun run build`
- `bunx tsc --noEmit`
- `bun run format:check`
- full bugreport91 end-to-end solve, route-completeness assertion, and final-board SVG snapshot
- necessary cramped branch-diversity and duplicate-explosion regressions
- SRJ15 reachability and e2e3 multisection tests
- bugreport66, bugreport78, and bugreport87 end-to-end regression tests
- full feature suite: 185 passed, 14 skipped, 0 failed
- remaining non-feature/non-bug suite: 247 passed, 22 skipped; its only failure reproduces unchanged on the untouched baseline
- legacy port selection compared exactly for SRJ15, SRJ18, bugreport66, bugreport78, and bugreport87


lib/solvers/NecessaryCrampedPortPointSolver/MultiTargetNecessaryCrampedPortPointSolver.ts
```diff
@@ -257,6 +299,62 @@ export class MultiTargetNecessaryCrampedPortPointSolver extends BaseSolver {
...
+      this.input.numberOfCrampedPortPointsToKeep + 2
+    return (
+      connectedTo.some((connectionId) =>
+        connectionId.startsWith("source_trace_"),
+      ) &&
+      !connectedTo.some((connectionId) =>
+        connectionId.startsWith("pcb_trace_"),
+      ) &&
+      !connectedTo.some((connectionId) =>
+        connectionId.startsWith("source_net_"),
```
seveibar: !!!!!!!!

seveibar (changes requested):
gotta look out for those hacks, extremely dangerous. Also notice that this is @0hmX 's territory

imrishabh18:
/benchmark --dataset srj20 --sample-numbers 126 --same-machine

0hmX (approved):

0hmX (changes requested):
Image

too much increase

0hmX:
my solution are also having the same problem it seems

PR #2132, opened 2026-08-17
Ignore zero-length preloaded trace sections

ShiboSoftwareDev:
## Stack

- Previous PR — real sample 11 PCB reproduction
- This PR — zero-length fixed-section fix

## Root cause and fix

`convertPreloadedTraceToHdRoutes` accepted a two-point route when both points had the same x, y, and layer. That route has no copper geometry, but downstream regional replacement still assigned it an identity and tried to reconnect around it.

The fix rejects only routes whose complete point list is the same 3D position. Real vias remain valid because their z values differ. There is no solver bypass, fallback, patch list, or DRC relaxation.

## Visual result

Before

After

The real PCB snapshots differ by 10.57% normalized pixel RMSE.

## Measured win

| Sample | Before | After |
| --- | --- | --- |
| 11 | reconnect corruption at 167.5 s | complete at 201.5 s |
| 19 | reconnect corruption at 874.7 s | corruption removed; autorouter timeout at 900 s |

The structural failure is removed in 2/2 affected samples and completion improves in 1/2. Sample 19 still needs a separate search-performance fix.

## Benchmark command

```text
/benchmark --pipeline 10 --dataset 29 --concurrency 4 --sample-timeout 900s
```

## Verification

- focused converter test: 1 passed, 0 failed, 4 expectations
- real sample 11 visual test: 1 passed, 0 failed, 6 expectations


ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --concurrency 4 --sample-timeout 900s

PR #2134, opened 2026-08-17
Use terminal layers to break stitch endpoint ties

ShiboSoftwareDev:
## Stack

- Previous PR — real sample 4 missing-via reproduction
- This PR — layer-aware endpoint tie-break fix

## Root cause and fix

Route stitching used 2D distance for endpoint selection and orientation. Colocated via endpoints therefore tied even when only one endpoint was on the requested terminal layer.

The fix uses layer agreement only to break equal-distance choices. It does not make other-layer endpoints infinitely far away and does not override a geometrically nearer endpoint. That narrower rule is important: the initial broad version solved sample 4 but made sample 10 time out; this version solves both.

## Visual result

Before

After

The focused real PCB detail differs by 1.63% normalized pixel RMSE and visibly restores the terminal transition.

## Measured win

| Metric | Baseline | Full root-fix stack |
| --- | ---: | ---: |
| Completed | 10/20 | 15/20 |
| Relaxed-DRC clean | 7/20 | 7/20 |
| Timeouts | 3/20 | 3/20 |
| P50 | 559.9 s | 204.1 s |
| P95 | 900.0 s | 900.0 s |

New completions are samples 3, 4, 6, 11, and 13. Sample 10 completes in 228.3 s with the narrow tie-break instead of regressing to a timeout.

## Benchmark command

```text
/benchmark --pipeline 10 --dataset 29 --concurrency 4 --sample-timeout 900s
```

## Verification

- focused cross-layer unit repro: 1 passed, 0 failed
- real sample 4 PCB visual: 1 passed, 0 failed, 7 expectations
- surrounding stitch and bug suite: 9 passed, 0 failed
- complete local benchmark used `--concurrency 1`


ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --concurrency 4 --sample-timeout 900s

PR #2137, opened 2026-08-17
Restore board-edge DRC in autorouting debugger

imrishabh18:
I want the `checkPcbTracesOutOfBoard` drc check to run, for that we need reconstruct the `pcb_board` from the SRJ


seveibar (approved):

seveibar:
@imrishabh18 DO NOT MERGE WITHOUT UPDATING BENCHMARK!

seveibar:
@imrishabh18 we need to make sure that the relaxed DRC is known and if we're decreasing it with this PR we understand why

PR #2143, opened 2026-08-18
Pin AM62L LPDDR4 repro and allow 8-layer Pipeline 10 inputs

ShiboSoftwareDev:
## Summary

- pin `@tscircuit/dataset-srj29-ddr3-bga-pairs` to `585c3abd4c6ce4be52bbc430b1d5e6dcc9b4ea30`
- add the exact 8-layer AM62L32-to-LPDDR4 sample as SRJ29 sample 021
- remove Pipeline 10's artificial 12-layer minimum so it uses the board's actual copper-layer count

## Why

The 12-layer check came from the original roomy SRJ29 fixture, not from a fanout-solver or Pipeline 9 requirement. It rejected the real 8-layer reproduction before fanout began.

## Validation

- focused Pipeline 10 tests: 3 passed, including a 19 ms 8-layer initialization check
- `bun run format:check`
- `bunx tsc --noEmit`
- sample 021 was not solved locally; the regression check uses a one-connection variant of the existing fast fixture

This PR supersedes #2142, whose deployment coincided with transient GitHub tarball `504` errors during dependency installation.


ShiboSoftwareDev:
/benchmark --dataset 29 --pipeline 10 --sample-timeout 900s

0hmX:
lgtm but will need review for @seveibar not sure why that limit was even present

ShiboSoftwareDev:
> lgtm but will need review for @seveibar not sure why that limit was even present

Wasn't needed, just AI slop

0hmX (approved):

PR #2144, opened 2026-08-18
Reproduce Pipeline 9 through-obstacle repair failure

ShiboSoftwareDev:
## What this reproduces

Dataset 29 sample 3 reaches Pipeline 9 joint DRC repair after both BGA fanouts. Pipeline 9 has materialized a changed fanout section as a `through_obstacle` trace, then rejects that same trace representation when exact repair starts.

The regression test intentionally records the current failure and snapshots the real PCB state using `circuit-to-svg`. The board has both BGA fanouts, but routing stops before the chips are connected.

PCB at the Pipeline 9 failure

## Scope

This PR contains only the reproduction test and its SVG snapshot. It does not change solver behavior.

## Validation

- `bun test tests/repro/pipeline9-through-obstacle-preloaded-trace.test.ts --timeout 9999999`
- `bun run build`

The fix will be submitted as a stacked PR so its PCB snapshot diff shows the completed routing.


ShiboSoftwareDev:
/benchmark --dataset 29 --pipeline 10 --sample-timeout 900s

PR #2145, opened 2026-08-18
Respect board edge clearance in RectDiff topology

imrishabh18:
Before
image

After
image

## Summary

- update RectDiff to preserve the physical board outline while applying `minBoardEdgeClearance` to its board-void topology blockers
- apply the constraint for every RectDiff caller, with no feature flag
- strengthen Bug Report 88 to assert zero total DRC errors and update affected route snapshots

The underlying RectDiff change is in https://github.com/tscircuit/rectdiff/pull/142.

## Validation

- `bun test tests/bugs/bugreport88-9a86ed.test.ts --timeout 9999999`
- affected 13-test snapshot/regression suite
- `bun run build`
- `git diff --check`
- GitHub CI: all 8 Bun Test shards, Build, Vercel Build, Type Check, Format Check, and No String ID Hacks

Bug Report 88 now produces zero total DRC errors.


seveibar (changes requested):
Never pass lies down the stack. Changing the outline by offsetting it creates more issues that it solves

seveibar:
You should find the underlying issue, and dont flag-gate it

PR #2146, opened 2026-08-18
Repair preloaded trace sections around through-obstacle connectivity

ShiboSoftwareDev:
Stacked on #2144.

## Root cause

Pipeline 9 can materialize fanout traces containing a `through_obstacle` primitive. That primitive records connectivity supplied by a connected component obstacle; it is not a physical via or wire that exact DRC repair may relocate.

Joint DRC repair previously required every selected preloaded trace to become one movable high-density route. It therefore rejected a selected trace containing `through_obstacle` and stopped samples such as dataset 29 samples 3 and 6.

Keeping the complete trace fixed avoids that exception but is also incorrect ownership: it unnecessarily freezes the ordinary wire and via geometry on both sides. The dataset benchmark exposed that coarse behavior as a completion regression from 50% to 20%, primarily through additional timeouts.

## Fix

Joint repair now splits a selected preloaded trace at each `through_obstacle` primitive:

- the connectivity primitive remains fixed;
- each ordinary copper section on either side becomes a separately anchored repair connection;
- DRC errors retain the identity of the section the portfolio can move;
- repaired sections are reassembled at their original route positions around the unchanged primitive.

This directly represents repair ownership. It does not catch an error, discard connectivity, synthesize a via, or add a fallback. Missing identities, positions, routes, and overlapping repaired ranges remain explicit solver invariant errors.

## PCB snapshot

The snapshot is generated from the real dataset 29 sample 3 PCB with `circuit-to-svg`.

| Repro: Pipeline 9 stops during joint repair | Fixed: ordinary sections route around preserved connectivity |
| --- | --- |
| failing PCB | fixed PCB |

The test also asserts that movable ranges exclude every original `through_obstacle` position and that the final routed PCB still contains the preserved primitive.

## Results

The [full 20-sample benchmark](https://github.com/tscircuit/tscircuit-autorouter/pull/2146#issuecomment-5329944670) uses the same Pipeline 10 command and 900-second sample limit as the repro benchmark:

| Dataset 29 result | Repro PR | Coarse trace-level ownership | Section-level ownership |
| --- | ---: | ---: | ---: |
| Completion | 50% (10/20) | 20% (4/20) | **60% (12/20)** |
| Relaxed DRC pass | 35% (7/20) | 15% (3/20) | **35% (7/20)** |
| Timeouts | 3 | 12 | **3** |

[...]

PR #2148, opened 2026-08-18
Allow benchmark-all to select a pipeline

ShiboSoftwareDev:
## What changed

`/benchmark-all` now accepts benchmark arguments such as `--pipeline 9` and forwards them to every configured dataset workflow.

The dispatcher appends the dataset selected by the fan-out to each argument list, so this command:

```text
/benchmark-all --pipeline 9 --same-machine
```

dispatches the default dataset plus srj18, srj19, srj20, srj21, and srj23. Each workflow benchmarks current main and the PR head sequentially on the same runner using Pipeline 9.

An explicit `--dataset` remains invalid for `/benchmark-all` because it conflicts with the command's dataset fan-out.

## Root cause

The parser previously rejected every `/benchmark-all` argument except `--same-machine`, and the dispatcher deliberately omitted `benchmark_args_json` for every fan-out run. Accepting the syntax alone would therefore still have run the default pipeline.

## Validation

- `bun test tests/pr-benchmark-command.test.ts --timeout 9999999`
- `bunx tsc --noEmit`
- `bun run build`
- Direct parser simulation verified all six generated argument lists contain `--pipeline 9` and their respective `--dataset`.

The issue-comment parser is loaded from the default branch, so the new command syntax becomes available after this PR merges.


ShiboSoftwareDev:
Moved this change to #2146 as commit 1d49502f8.

PR #2154, opened 2026-08-18
Fix bugreport93 overlapping via clearance

seveibar:
## Summary

- update high-density-repair03 to tscircuit/high-density-repair03#73
- ensure exact targeted repair prioritizes different-net via clearance before trace-topology candidates can consume the iteration budget
- fix the overlapping via pair reproduced and snapshotted in #2150

## Validation

- replayed the complete captured routing through the pinned dependency: the reported pair is removed and zero via-clearance errors remain
- bun test tests/bugs/bugreport93-overlapping-vias.test.ts (skipped as intended)
- bunx tsc --noEmit
- bun run build

Depends on tscircuit/high-density-repair03#73.


seveibar:
The repair scheduling is now bounded: same-net via errors remain immediate, while different-net via errors receive one targeted slot every eight iterations instead of preempting trace topology every iteration. This keeps the bugreport93 fix and removes the benchmark penalty. Updated same-machine SRJ18 result: P50 -2.5%, P60 +1.7%, P70 -6.5%, P80 -7.9%, P90 -6.3%, one fewer timeout, and no outcome regressions.

PR #2159, opened 2026-08-19
Fix Map iterator compatibility in IntraNodeRouteSolver

### LINKED ISSUE #2106: Autorouting Bug Report: <board#104 />
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=bb9c95e2-8736-4391-b95f-7120887245ec

(any additional details? Maybe a screenshot?)

This is just the new project template/example.

This error was shown

Image

```
o2.entries().map is not a function. (In 'o2.entries().map(([t53,e7])=>({connectionName:t53,rootConnectionName:this.rootConnectionNameByConnectionName.get(t53),points:Vi2(e7)}))', 'o2.entries().map' is undefined) (capacity-autorouter@0.0.802)
```

obmakesomething:
## Summary

- convert `Map.entries()` to an array before calling `.map()`
- add a regression test that simulates browser runtimes without Iterator Helpers
- verify that connection names and points are preserved during solver construction

## Root cause

`Map.prototype.entries()` returns an iterator. Calling `.map()` directly on that iterator depends on the newer Iterator Helpers proposal, which is not available in every browser/runtime supported by the package. The constructor therefore throws before routing starts.

The safe pattern is already used immediately above this code path: `Array.from(map.entries()).map(...)`.

## Test plan

- `bun test tests/bugs/issue2106-map-iterator-compat.test.ts --timeout 9999999`
- `bun run build`

Fixes #2106


obmakesomething:
The current head (`12034aa`) has all PR workflows in `action_required` with zero jobs, including the Autorouting Benchmark run (`32247876029`), so `/benchmark` could not produce a result. Could a maintainer approve the fork workflow runs when convenient? I’m not treating the absent jobs as a code or benchmark failure.

obmakesomething:
@seveibar This PR is ready for maintainer review. The upstream checks are currently in `action_required` because the workflow runs for this fork PR still require maintainer approval.

The change is intentionally narrow:
- materialize `Map.entries()` with `Array.from(...)` before `.map()`;
- preserve connection names and points;
- add a regression test that disables Iterator Helpers to reproduce issue #2106;
- no routing algorithm or geometry behavior changes.

Local verification completed with the focused regression test and package build, and the Vercel preview is ready. Could you approve the workflow runs or let me know what additional validation you want?

Issue #74, opened 2025-03-21
HyperCapacityPathing to fix capacity pathing issues


seveibar:
We currently order traces routed in the capacity pathing stage by how they were provided in the SimpleRouteJson. For the HyperCapacityPathing, we'll create sets of hyperparameters similar to the HyperHighDensitySolver that determine ordering as well as how greedy the algorithm should be

Currently the algorithm is extremely greedy and can gobble too many small nodes that are needed (especially for "escape routing" where a pad may be surrounded by small nodes), but small adjustments to the cost function can prioritize larger nodes. We can run multiple solvers with different hyper parameters with Hyper A* to allow more complicated pathing solutions to run (ones that don't eliminate the solution space as quickly- e.g. some could even allow traversing through nodes that have fully exhausted their capacity)

The main hyperparameter we need to adjust is how much to subsidize traversal through large nodes. Boards that have difficult escape situations will want to prioritize through large nodes (escaping to an "open area") rather than using up all the small nodes. There are a lot of "rules of thumb" that can be encoded weighted with hyperparameters

seveibar:
```
I'm trying to create a new capacity pathing stage. The old stage can fail because nodes get "boxed in", so I'm trying a new approach.

The new stage, which I call a CapacityPathingMultiSectionSolver, runs a greedy pathing solver that is "unconstrained", it allows nodes to go into negative remaining capacity.

We're then going to look for these problematic nodes, and create a section around them by doing a breadth first search to N degrees (e.g. we look up to 6 nodes away)

We then "re-resolve" the section with a CapacityPathingSingleSectionSolver. The HyperSingleSectionSolver will create different CapacityPathingSingleSectionSolvers with different SHUFFLE_SEED to attempt to create a pathing that doesn't have nodes that exceed their capacity
```

PR #2164, opened 2026-08-19
Eliminate Pipeline7’s remaining dataset01 DRC failure

AnasSarkiz:
## Impact

Targets the remaining Pipeline7 dataset01 DRC failure after #2162: **84/85 → 85/85 (100%)**.

The remaining board is dataset01 circuit143 (sample58), where the final output has a real trace-to-trace clearance gap of about **0.015 mm**. Pipeline9 is unchanged.

## Root cause and fix

The first targeted repair reduces circuit143 from two DRC errors to one, but its trace move exposes a new different-net trace conflict. Clearing that last conflict requires one coupled candidate: a geometry-derived trace detour plus relocation of the newly blocking movable via.

- Pin [high-density-repair03#74](https://github.com/tscircuit/high-density-repair03/pull/74), which implements and fully scores that generic coupled repair.
- Run one bounded exact DRC portfolio directly in Pipeline7.
- Delete the adaptive probe/retry wrapper and its swallowed-error path.
- Keep broad repair disabled; broad-only settings and geometry are never initialized.
- Add circuit143 and no-fallback regressions, plus strengthen the existing dataset18 exact-repair regression.

There are no circuit IDs, fixed coordinates, endpoint normalization, DRC filtering, or fallback routing in the production fix.

## Validation

- `bun run build`
- Focused repair suite: **3 passed, 0 failed**
- Complete Pipeline7 suite: **36 passed, 1 existing skip, 0 failed**
- circuit143 final relaxed DRC: **0 errors**
- Broad repair: **not initialized or attempted**
- Official dataset01 benchmark: pending PR run

## Dependency

Pinned to the exact upstream head of [high-density-repair03#74](https://github.com/tscircuit/high-density-repair03/pull/74): `5eb78358e7910c1095bc80edeb677360183080fa`.

Before this PR merges, the pin will be updated to #74's eventual upstream merge commit.


AnasSarkiz:
Superseded by #2173. Autorouter `main` advanced with the bugreport93 repair03 pin after this PR opened, so the old dependency head would drop that newer fix. #2173 is rebuilt from current `main`, pins the combined clean repair03 PR #75, and remains exactly one commit. Closing this stale PR without force-pushing its history.

PR #2169, opened 2026-08-19
Reproduce fanout continuation reserving a routing region

ShiboSoftwareDev:
## What this reproduces

SRJ29 sample007 reaches Pipeline 10 with two real fanned-out BGAs, but tiny-hypergraph port pathing treats a fanout continuation point as if it owned the entire containing routing region.

With a fixed 250,000-iteration budget on `main`, pathing exhausts the budget and never connects the two fanout stubs.

This PR changes no solver behavior.

## Real PCB snapshot

The SVG is generated with `circuit-to-svg` from the real sample007 pads and solver-produced fanout copper. It highlights one affected DDR3 BA0 net so the missing connection is visible without a misleading all-layer overlay.

Before: disconnected BA0 fanout stubs

## Validation

- `bun test tests/repro/pipeline10-srj29-terminal-region-ownership.test.ts --timeout 300000`

The fix remains isolated in the next PR in the stack.


ShiboSoftwareDev:

ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 18 --sample-timeout 900s

PR #2170, opened 2026-08-19
Fix already-connected endpoints reserving shared routing regions

ShiboSoftwareDev:
Stacked on #2169.

## Root cause

The tiny-hypergraph adapter inferred that every capacity region containing a connection endpoint belonged exclusively to that endpoint's net.

That is correct for a desired connection endpoint inside its terminal region, but not for an endpoint that is already connected by fanout copper and passed to the next routing stage. The latter only marks where the remaining route begins; owning its large containing region turns shared routing space into a topology barrier across every available layer. Reripping cannot repair a missing graph path.

## Fix

- Desired connection endpoints with `pcb_port_id` keep region ownership.
- Legacy `_containsTarget` desired connection endpoints keep region ownership.
- Regions associated with already-connected copper through `_connectedTo` keep region ownership.
- Endpoints already connected to fanout copper explicitly leave their ordinary containing routing region shared (`netId: -1`), preventing the tiny-hypergraph loader from inferring exclusivity again.

There is no dataset-name check, point-ID prefix check, fallback, solver-order change, or larger iteration budget.

## Real PCB result

Before, the highlighted BA0 net stops at two disconnected fanout stubs:

Before: disconnected BA0 fanout stubs

After, the same test runs the normal Pipeline 9 pathing, high-density-routing, and stitching stages, then renders the actual connected solver output with `circuit-to-svg`:

After: connected BA0 route

The highlighted route has zero relaxed-DRC errors.

## Evidence

| Check | Reproduction PR | This PR |
| --- | ---: | ---: |
| Port pathing at 250,000-iteration budget | Exhausted | Solved |
| High-density routing and stitching | Not reached | Completed |
| Highlighted BA0 route relaxed DRC | N/A | 0 errors |

Ten legacy tests that failed under the previous attempted fix now pass with their existing snapshots unchanged, including Pipeline 4, Pipeline 7, SRJ15, SRJ18, SRJ19, SRJ20, bugreport68, bugreport69, bugreport73, bugreport74, and bugreport75 coverage.

The full sample still has eight later pad-to-trace clearance errors from fanout/routing quality. This PR fixes the topological pathing failure only and does not claim that separate issue is solved.

## Validation

- `bun test tests/repro/pipeline10-srj29-terminal-region-ownership.test.ts --timeout 300000`
- Focused legacy tests run one file at a time
- `bunx biome check lib/solvers/PortPointPathingSolver/tinyhypergraph/getRegionNetIdByRegionId.ts tests/repro/pipeline10-srj29-terminal-region-ownership.test.ts`
- `bun run build`


ShiboSoftwareDev:

ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --sample-timeout 900s --same-machine

lib/solvers/PortPointPathingSolver/tinyhypergraph/getRegionNetIdByRegionId.ts
```diff
@@ -59,6 +67,13 @@ export function getRegionNetIdByRegionId(input: {
   }
 
   const regionNetIdByRegionId = new Map<string, number>()
+  for (const regionId of logicalContinuationRegionIds) {
+    // A non-PCB continuation endpoint inside ordinary routing space does not
+    // make that whole region exclusive. Serialize -1 explicitly so the loader
+    // does not infer ownership again from connection start/end regions.
+    // Physical PCB terminals and connected-trace regions are assigned below.
```
seveibar: huge anti-pattern that keeps coming up.

avoid using "physical" and "logical", nobody knows what these mean

use "alreadyConnected" and "desiredConnection" etc., that is much more clear

seveibar (approved):

lib/solvers/PortPointPathingSolver/tinyhypergraph/getRegionNetIdByRegionId.ts
ShiboSoftwareDev: Updated this to use intent-based names: `alreadyConnectedEndpointRegionIds` and `isDesiredConnectionEndpoint`. The comment now explains that an endpoint already connected to copper only marks where desired routing resumes, while desired-connection endpoints and connected-copper regions retain ownership. This is a terminology-only change; the focused SRJ29 reproduction and build still pass.

PR #2172, opened 2026-08-19
Add no-breakout hierarchical routing failure repro

Abse2001:
## Summary

- capture the real child and parent routing inputs from the RP2350 hierarchical board
- route the child subcircuit first, then hand its generated traces to the parent router
- reproduce the default parent router exhausting its iterations
- add an uncropped snapshot of the successful child result and full parent input

## Bug

The child subcircuit routes its 22 internal connections into 70 traces with zero relaxed-DRC errors. Those freshly generated traces are then preserved in the 62-connection parent input. The default parent autorouter exhausts its iterations instead of completing the remaining board routes.

There are no breakout points, manually constructed output routes, hard-coded route paths, custom autorouter options, or cropped/zoomed snapshots in this repro.

## Test

- `bunx tsc --noEmit`
- `BUN_UPDATE_SNAPSHOTS=1 bun test tests/bugs/subcircuit-then-parent-route-failure.test.ts --timeout 9999999`


Abse2001:
Closing in favor of a Core-level JSX reproduction, following tscircuit/core#3298 so Core itself constructs and routes the child subcircuit before the parent phase.

PR #2181, opened 2026-08-20
Reproduce Pipeline 9 zero-length preloaded section reconnect failure

ShiboSoftwareDev:
## Problem

Pipeline 9 fails while reconnecting SRJ29 sample 011 after BGA fanout:

```text
Pipeline9 could not reconnect mutated preloaded segment "DDR3_a3_i_mx6ull_main_ram_ddr3_dram_data15_fixed_27_1"
```

The affected DQ15 fanout contains a preloaded wire section whose two endpoints have the same position and layer. It has no copper extent, but Pipeline 9 gives it a route identity and later attempts to reconnect geometry that does not exist.

## Clear visual reproduction

This snapshot renders only the target net that Pipeline 9 successfully materialized, using real solver output and `circuit-to-svg`:

Materialized DQ15 output before the fix

The routed middle exists, but the two real BGA pads are visibly disconnected because Pipeline 9 throws before it can reconstruct the fanout trace. No trace is manually removed from a successful result: the snapshot includes the fanout only when `getUpdatedPreloadedTraces()` successfully materializes it.

The test also keeps the full-board DQ15 snapshot for context:

Full DQ15 route at the reconnect failure

## Reproduction

- Uses the real 18-layer SRJ29 DDR3/BGA sample 011.
- Runs Pipeline 10 fanout and Pipeline 9 through high-density stitching.
- Asserts the exact failing preloaded-section name.
- Produces both PCB SVGs from actual circuit JSON.

This PR contains only the reproduction and generated SVG snapshots. It does not change solver behavior.

## Validation

```sh
bun test tests/repro/pipeline9-srj29-zero-length-preloaded-section.test.ts --timeout 9999999
```

Pipeline 10 benchmark (`--dataset 29 --sample 11`): failed at the reproduced reconnect error after 136.0s.

Stacked fix: #2182


ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --sample 11 --sample-timeout 900s

ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --sample 11 --sample-timeout 900s

PR #2182, opened 2026-08-20
Omit zero-length Pipeline 9 preloaded trace sections

ShiboSoftwareDev:
## Fix

Stacked on reproduction PR #2181.

Do not create a high-density route for a preloaded trace section when every point has the same `x`, `y`, and layer index.

That section represents no copper. Previously it still received a route identity, so the reconnect stage later searched for geometry that did not exist. The check is applied directly at the conversion boundary where preloaded sections become high-density routes.

Real vias remain valid because their layer index changes even when `x` and `y` are identical.

This does not catch the reconnect error, add a fallback, relax DRC, increase a search budget, or reorder the pipeline.

## Visual result

Both images are generated from the same real DQ15 net with `circuit-to-svg`. They show only the output Pipeline 9 successfully materializes.

| Before | After |
| --- | --- |
| Disconnected materialized output | Connected materialized output |

Before the fix, the routed middle is stranded from both BGA pads because reconstruction throws. After the fix, the red fanout at the upper-left and the green fanout at the lower-right visibly join those pads to the routed middle.

## Validation

```sh
bun test tests/features/pipeline9-skips-zero-length-preloaded-sections.test.ts --timeout 9999999
bun test tests/repro/pipeline9-srj29-zero-length-preloaded-section.test.ts --timeout 9999999
```

The real sample reaches `traceSimplificationSolver` after high-density route stitching without the reconnect exception.

## Benchmark

Pipeline 10, SRJ29 sample 011, 900s sample timeout:

| Revision | Result | Time | Last state |
| --- | --- | ---: | --- |
| Reproduction #2181 | Failed | 136.0s | Zero-length DQ15 section reconnect error |
| This fix | Timeout | 900.0s | 75% pipeline progress, 594,581 iterations |

The fix removes the deterministic reconnect failure and lets the sample enter later autorouting work. It does not make sample 011 complete; the later timeout is a separate solver-performance issue.


ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --sample 11 --sample-timeout 900s

ShiboSoftwareDev:
/benchmark --pipeline 10 --dataset 29 --sample 11 --sample-timeout 900s

ShiboSoftwareDev:

PR #2193, opened 2026-08-21
Fix Pipeline 9 when regional routing merges fixed sections

Abse2001:
## Summary

- preserve an original fixed section when a grid-based regional solver merges it with electrically equivalent same-net copper and emits no separately named replacement
- include every preserved section in candidate copper-conflict validation before accepting the regional result
- count only sections that were actually replaced in rerouting statistics
- regenerate the full no-breakout Game Boy input from `tscircuit@0.0.2384` (`@tscircuit/core@0.0.1730`, autorouter `0.0.830`)
- convert the refreshed full no-breakout Game Boy regression to a passing routed snapshot

Depends on #2192. Until that repro-only PR merges, this stacked PR also displays its fixture commits. After #2192 merges, this PR will reduce to the fix commits.

## Root cause

The full Game Boy repro reaches a small regional node containing two almost coincident fixed sections on the same root net. HighDensity A01/A03 rasterize both sections into the same grid cells and intentionally emit one physical route. Pipeline 9 previously required exactly one named replacement for every movable fixed section and failed with `expected one replacement ... got 0` even though the omitted original copper could remain safely in place.

The change does not discard or invent copper. A section with no replacement remains unchanged and is added back to the fixed-route collision set. A candidate that conflicts with it still fails or retries through the existing explicit promotion path. Multiple replacements remain an invariant violation.

## Result

The refreshed 21-connection, 379-obstacle, 136-preloaded-trace Game Boy parent board now routes without breakout points in about 179 seconds. The routed output contains 140 traces and 153 vias. It still has 32 relaxed-DRC findings, so this PR fixes completion rather than claiming DRC-clean routing.

Full Game Boy routed with Pipeline 9

## Validation

- `bun test --timeout 9999999 tests/bugs/bugreport96-full-gameboy-no-breakout.test.ts`
- `bun test --timeout 9999999 tests/features/pipeline9-high-density-regional-fallback.test.ts tests/features/pipeline9-high-density-no-invalid-fallback.test.ts tests/features/pipeline9-regional-fallback-target-layers.test.ts tests/features/pipeline9-regular-regional-fallback-fixed-preload.test.ts`
- `bunx tsc --noEmit`
- `bun run build`
- `bun run format:check`
- `git diff --check`


seveibar (approved):
i think for Pipeline9 tests we need a horizontally stacked before/after snapshot

seveibar:
this may cause DRC failures? Not sure

PR #2199, opened 2026-08-22
feat: constrain autorouting to selected layers

seveibar:
## Summary

- add `routingLayers?: string[]` to Simple Route JSON
- constrain newly generated wire segments and logical layer transitions across Pipelines 1–10 and the assignable pipelines
- intersect multilayer terminals, buses, capacity nodes, polygraph regions, fanout escapes, copper-pour escapes, and repair candidates with the allowlist
- preserve existing/preloaded traces, including caller-authored copper on excluded layers
- reject empty, duplicate, unknown, single-layer-incompatible, or terminal-incompatible selections with actionable errors
- keep fallback, jumper, power-expansion, and differential-pair stages from silently reopening excluded layers

## Why

Multilayer boards often reserve inner copper for power/ground planes. The autorouter currently treats every board layer as a signal-routing candidate, so a board can silently fragment those planes with unrelated traces.

This is the router enforcement for the board API in tscircuit/props#808 and the transport fields in tscircuit/circuit-json#719. Omitting `routingLayers` preserves the current all-layer behavior.

`routingLayers` controls wire layers and logical routing transitions. The physical plated span of a via is intentionally independent and is controlled by `viaSpanPolicy`; for example, an ordinary through-via can cross a reserved plane without routing a wire on that plane.

## Unsupported external moves

Some optional postprocessors currently enumerate every board layer and do not accept an allowlist. When a proper subset is selected:

- unrestricted high-density repair phases are disabled while safe repair/simplification phases remain enabled
- differential-pair length-matching is a no-op only when no pairs were requested; requested pair work fails closed until the external solver accepts an allowlist
- power expansion fails closed if it is allowed to add vias
- the top-only jumper stage fails closed when a crossing would require a top-layer jumper but `top` is excluded

This prevents a late fallback or optimization stage from escaping onto a reserved layer.

## Validation

- 17 focused routing-layer files pass with 106 assertions
- coverage includes every routing pipeline family, Pipeline 10 fanout, P4/P9 crossing fallbacks, power-expansion defaults, Assignable Pipeline 3 jumpers, differential-pair postprocessing, top-only and top/bottom four-layer output, one-layer validation, terminals/buses, pour escapes, preloaded traces, and repair policy
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`
- broad suite run completed the functional/routing coverage; the only observed SVG snapshot mismatch (`bugreport51-7db9f8`) reproduces unchanged on current `main` in the same environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ The input to the autorouter is a `SimpleRouteJson` object with the following str
 ```typescript
 interface SimpleRouteJson {
   layerCount: number
+  routingLayers?: string[]
   minTraceWidth: number
   obstacles: Obstacle[]
   connections: Array<SimpleRouteConnection>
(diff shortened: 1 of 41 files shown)


seveibar:
Closing after API review: board routing intent will use the existing <copperpour unbroken /> / isCopperPour SRJ obstacles rather than a separate routingLayers allowlist. These obstacles spatially reserve the pour for foreign-net wires while still permitting through-via antipads and same-net plane escapes. Normal obstacles remain hard keepouts. The quick-config failure came from its plane pours not setting unbroken. Blind/buried fabrication control continues in #2200 as allowBlindAndBuriedVias.

Issue #2156, opened 2026-08-18
Default 4-layer routing emits blind/buried vias without an opt-in or fabrication constraint


0hmX:
## Summary

The default multilayer autorouting path emits partial-stack vias (blind/buried layer pairs) even when the board does not request or opt into blind/buried-via fabrication.

For a normal `<board layers={4}>` with only pad/drill size rules, I expected ordinary full-stack PTH vias unless allowed via spans were explicitly supplied. Instead, the route used F.Cu–In1, F.Cu–In2, In1–B.Cu and In2–B.Cu vias.

This is an unsafe default: many mainstream fabrication profiles, including JLCPCB, do not support blind/buried vias. It also exposed [tscircuit/cli#4346](https://github.com/tscircuit/cli/issues/4346), where non-through drill programs are omitted from the Gerber ZIP.

## Reproduction

Environment:

- `tscircuit@0.0.2356`
- `@tscircuit/capacity-autorouter@0.0.815`
- 4-layer, 1.6 mm board
- no via-type or allowed-layer-pair configuration
- `minViaHoleDiameter="0.2mm"`
- `minViaPadDiameter="0.35mm"`

Command:

```sh
bunx tsci build index.circuit.tsx --autorouter-timeout 30s
```

The route completed successfully with 65 traces, 0 jumpers and 56 vias. Via spans in the emitted Circuit JSON were:

| Span | Count |
|---|---:|
| F.Cu–B.Cu | 25 |
| F.Cu–In1 | 13 |
| F.Cu–In2 | 6 |
| In1–B.Cu | 10 |
| In2–B.Cu | 2 |

Representative emitted record:

```json
{
  "type": "pcb_via",
  "x": -4.002,
  "y": -8.306,
  "hole_diameter": 0.2,
  "outer_diameter": 0.35,
  "layers": ["top", "inner1"],
  "from_layer": "top",
  "to_layer": "inner1"
}
```

No warning or manufacturing-capability error was produced.

## Expected behavior

Default routing should use full-stack F.Cu–B.Cu PTH vias.

Blind, buried and microvia layer pairs should be available only when explicitly enabled by a board/fabrication capability such as an allowed-via-span list or stackup definition.

At minimum, the router/build should report that advanced via types were introduced.

## Suggested design

- Add allowed via spans to SimpleRouteJson/autorouter options.
- Default a conventional multilayer board to `top -> bottom` only.
[...]

Issue #79, opened 2025-03-26
`keyboard4` DRC issue with close traces due to trace simplification stage


seveibar:
![Image](https://github.com/user-attachments/assets/90e974a3-e335-4c00-acdf-47493137bcf3)

seveibar:
fixed

PR #2200, opened 2026-08-22
feat: add blind and buried via control

seveibar:
## Summary

Adds `allowBlindAndBuriedVias?: boolean` to Simple Route JSON and enforces the fabrication-safe setting across every native autorouter output path.

- `allowBlindAndBuriedVias: false` emits each newly generated or moved via as a physical `top`-to-`bottom` plated barrel. Logical routing may still enter or leave that barrel on an inner layer.
- Omission or `allowBlindAndBuriedVias: true` preserves the standalone autorouter's legacy blind/buried-via behavior.
- Ordinary tscircuit boards default this value to `false` through the companion props/core plumbing. HDI callers that intentionally require blind or buried vias can opt in explicitly.

Addresses #2156.

## Routing and safety behavior

The control is applied at all supported native output boundaries: Assignable Pipelines 1/2/3 and Pipelines 1/2/3/4/6/7/8/9/10.

When blind and buried vias are disabled, generated vias are materialized and validated across their full inclusive physical stack before output. A `top`-to-`bottom` barrel occupies `top`, every `innerN`, and `bottom`; routing fails closed if it intersects unrelated copper, an axis-aligned or rotated keepout, a serialized route jumper pad, or an authored `srj.jumpers[].pads` pad on any crossed layer. Validation includes non-superseded input copper as well as newly routed traces.

The two external generation paths are covered too:

- Pipeline 10 materializes and validates fanout-solver vias before feeding fanout output into Pipeline 9 as preloads. Both `getOutputSimpleRouteJson()` and the standard `getOutput()` expose only solved, validated output.
- Power-trace expansion materializes and validates newly inserted vias, including when `allowNewVias: true` is requested.

Unchanged vias supplied in input/preloaded traces keep their authored spans and exact serialized shape. Pipeline 9 tracks fixed-route and per-via provenance, restores authored via objects, and materializes only generated replacements. This prevents synthesized defaults such as a hole diameter from reclassifying an untouched partial via.

Same-net pads and inner-layer taps remain legal.

### Copper pours

The board-level `<copperpour unbroken>` choice is emitted to the router as `isCopperPour` obstacles. Those plane obstacles still reserve their region against foreign-net wire routing. The generated-via validator alone does not treat them as hard via keepouts because a plated barrel may cross a plane and receives the appropriate connection or clearance antipad. A normal obstacle remains a hard keepout for both wires and via barrels. Downstream pour generation and final Circuit JSON/Gerber checks remain responsible for verifying the resulting plane connection or antipad.

## Compatibility and companion work

Standalone autorouter behavior is unchanged when `allowBlindAndBuriedVias` is omitted or `true`; `false` enables the new fail-closed behavior. Invalid runtime values are rejected instead of being silently interpreted.

Disabling blind and buried vias can make dense HDI/BGA fanout problems unsatisfiable when the geometry is only feasible with partial-depth vias. In that case the pipeline fails closed instead of returning unsafe fabrication output; such callers must explicitly set `allowBlindAndBuriedVias: true`.

The board-facing default and transport are provided by:

- tscircuit/props#808
- tscircuit/circuit-json#719
- tscircuit/core#3374

Defense-in-depth companions can land independently:

- tscircuit/high-density-repair03#79 indexes a physical via on every crossed layer so conflicts can be repaired earlier.
- tscircuit/checks#233 validates plated spans in rendered Circuit JSON.
- tscircuit/cli#4354 covers the corresponding drill-export behavior.

This PR intentionally keeps the current repair-engine dependency. Its own final-output validation makes `allowBlindAndBuriedVias: false` fail closed without #79; once #79 is consumed, some intermediate-layer conflicts can be detected and repaired earlier rather than rejected at the output boundary.

## Validation

```text
bun test \
  tests/features/allow-blind-and-buried-vias.test.ts \
  tests/features/disallow-blind-buried-vias-all-native-output-paths.test.ts \
  tests/features/disallow-blind-buried-vias-collision-fail-closed.test.ts \
  tests/features/disallow-blind-buried-vias-jumper-pad-collision.test.ts \
  tests/features/disallow-blind-buried-vias-preserved-input-collision.test.ts \
  tests/features/disallow-blind-buried-vias-rotated-obstacle-collision.test.ts \
  tests/features/pipeline10-disallow-blind-buried-vias-fail-closed.test.ts \
  tests/features/pipeline9-disallow-blind-buried-vias-preserves-authored-vias.test.ts \
  tests/features/power-trace-expansion-disallow-blind-buried-vias.test.ts \
  tests/power-trace-expansion-empty-selection.test.ts
# 10 pass, 0 fail, 52 expect() calls (plain Bun defaults; no test timeout overrides)

bun test --timeout 30000 \
  tests/features/pipeline9-srj23-sample21-regional-b01-repair.test.ts \
  tests/bugs/bugreport-b5b3b9d8-drc-count.test.ts
# 2 pass, 0 fail, 11 expect() calls

bunx tsc --noEmit
# pass

bun run build
# pass

git diff --check
# pass
```

The regression matrix covers physical full-stack output, omission/explicit-`true` compatibility, terminal vias, every native and Assignable output boundary, rotated and axis-aligned keepouts, copper-pour antipads, same-net copper, both jumper-pad representations, preserved caller vias and copper, power-expander vias, and Pipeline 10 fanout fail-closed behavior.

## Credit

This is a self-contained current-main implementation of the API direction originally explored and benchmarked in #2117 by @ShiboSoftwareDev. The commit retains ShiboSoftwareDev's co-author trailer.


seveibar:
The remaining bugreport90 snapshot delta is expected from the companion full-span via DRC dependency in tscircuit/high-density-repair03#79. I isolated it on autorouter main (4c921fe2) with no feature changes and only the high-density-repair03 SHA changed; that produced the same snapshot mismatch. Commit 24811bab updates the visual baseline, and the focused snapshot test passes locally. This is not an omitted/unrestricted viaSpanPolicy compatibility change.

seveibar:
Follow-up to the earlier snapshot note: I separated the high-density-repair03 #79 dependency migration from this PR. The widespread legacy snapshot/stat changes were caused solely by that dependency bump, not by viaSpanPolicy. Head 5ef9244f restores the existing repair dependency and baseline snapshot, fixes the one actual compatibility regression (empty power-selection output array identity), and keeps through_only safe via the policy's own full-span final-output validator. Independent audit confirmed removing #79 can only make a conflict fail later; it cannot emit unsafe output. Focused compatibility/policy suite: 12/12, 63 assertions; representative legacy snapshots, typecheck, and build pass.

PR #2206, opened 2026-08-23
Update A03 to preserve terminal layer transitions

seveibar:
## Summary

- update the pinned high-density A03 solver to the terminal-layer fix from tscircuit/high-density-a01#89
- keep Pipeline9's explicit-via invariant intact instead of restoring a Core-side workaround
- fix the tscircuit/core#3378 Arduino center-region regression where a route ended on another layer without a via

## Dependency

This is stacked on tscircuit/high-density-a01#89. After that PR is squash-merged, this pin should be updated to its merged commit before this PR is merged.

## Verification

- captured Core Pipeline9 input: solved 17 connections, 184679 iterations, no missing-via error
- clean Core center-region test with rebuilt linked packages: 1 pass, 124 assertions
- bun run build
- bunx tsc --noEmit
- bun run format:check
- git diff --check
- full autorouter suite: 567 pass, 59 skip; 7 visual snapshot mismatches also reproduce unchanged on the baseline high-density commit and are unrelated to this dependency bump


seveibar:
Superseded by #2212, which fixes the Pipeline9 input contract and avoids sending interior splice anchors to A01/A03. The high-density-a01 dependency workaround is no longer needed.

PR #2215, opened 2026-08-24
Reproduce Pipeline 9 MangoPi pre-power connectivity loss

AnasSarkiz:
## Summary

- add the exact phase-free six-layer MangoPi R3C whole-board SRJ
- add the exact `powerTraceExpansionSolver` constructor tuple captured at the Pipeline 9 pre-power boundary
- add an explicit `AutoroutingPipelineSolver9_PreloadedTraceGraph` debugger fixture
- independently validate original endpoint connectivity from physical wire and via geometry
- keep the 1,122-second whole-pipeline reproduction opt-in and make no solver implementation changes

## Reproduced result

The input contains 113 connections and 518 endpoints. The captured pre-power output contains 405 traces and 519 via occurrences, and all 113 connection names are represented.

Physical-copper validation reports only 107 connected connections and 502 connected endpoints. The disconnected endpoint counts are:

| Connection | Disconnected endpoints |
| --- | ---: |
| `source_net_0` | 4 |
| `source_net_1` | 4 |
| `source_net_3` | 2 |
| `source_net_9` | 1 |
| `source_net_36` | 3 |
| `source_net_101` | 2 |

The captured run reached `powerTraceExpansionSolver` and later reported the whole pipeline as solved. The pre-power output is invalid and is not fabrication-ready.

## Scope

This is reproduction-only. It does not change Pipeline 9 or power-expander implementation, modify either SRJ artifact, alter routing constraints, update dependencies, or bundle DRC, USB, board-edge, or via-policy changes.

## Validation

- `bun test tests/bugs/bugreport99-mangopi-r3c-pipeline9-connectivity.test.ts --timeout 9999999` — 1 pass
- default manual test invocation — 1 skip, 0 fail
- `bunx tsc --noEmit` — pass
- `bun run build` — pass
- `bun run format:check` — pass
- exact input and constructor-tuple SHA-256 checks match the captured report

Run the full solve manually with:

```sh
RUN_BUGREPORT99_MANGOPI_PIPELINE9=1 \
  bun test tests/bugs/bugreport99-mangopi-r3c-pipeline9-connectivity.manual.test.ts \
  --timeout 9999999
```

PR #2216, opened 2026-08-24
Fix cross-layer stitch endpoint selection

AnasSarkiz:
## Stack

- Base: #2215 — reproduces the MangoPi Pipeline 9 pre-power connectivity loss
- This PR — fixes cross-layer endpoint selection during route stitching

## Root cause

Route stitching broke equal-distance endpoint ties with coordinate ordering alone. When route islands met at the same XY position on different copper layers, stitching could select the wrong-layer endpoint, omit the island containing the required via, and leave original terminals physically disconnected even though Pipeline 9 later reported solved.

## Fix

- Prefer terminal-layer agreement when endpoint distances tie.
- Preserve layer-aware orientation when assigning stitch start and end terminals.
- Do not snap an island endpoint across layers without retaining its transition.
- Add a focused regression proving that a colocated cross-layer terminal keeps its via-bearing route island.

This revives and adapts the approach from the stale, automatically closed #2134 to the current Pipeline 9 stitch safeguards.

## Validation

- `bun run format`
- `bun test tests/stitch-solver --timeout 9999999` — 9 passed
- `bun test tests/bugs/bugreport99-mangopi-r3c-pipeline9-connectivity.test.ts --timeout 9999999` — 1 passed
- `bun run build`

### DIFF
--- a/lib/solvers/RouteStitchingSolver/MultipleHighDensityRouteStitchSolver3.ts
+++ b/lib/solvers/RouteStitchingSolver/MultipleHighDensityRouteStitchSolver3.ts
@@ -19,6 +19,7 @@ import {
 } from "./routeStitchingEndpointHelpers"
 import {
   compareRoutes,
+  DISTANCE_TIE_TOLERANCE,
   MAX_TERMINAL_STITCH_GAP_DISTANCE_3,
 } from "./routeStitchingShared"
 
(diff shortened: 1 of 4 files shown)


AnasSarkiz:
Full MangoPi pre-power validation disproved the current fix. With the manual test temporarily unskipped, Pipeline 9 still produced exactly 107/113 connected connections and 502/518 connected original endpoints, with the same 16 endpoints disconnected across source_net_0, source_net_1, source_net_3, source_net_9, source_net_36, and source_net_101.

The CI failures are SVG snapshot changes caused by the over-broad stitch tie-break; those snapshots should not be updated. I have marked this PR draft while the actual stage that introduces the connectivity loss is isolated.

PR #2218, opened 2026-08-24
Fix preloaded trace parent net DRC attribution

Abse2001:
## Summary\n\n- keep explicit PCB port metadata authoritative when reconstructing pads for DRC\n- resolve preserved subcircuit trace aliases through the original parent connectivity graph\n- keep newly generated parent-route net resolution unchanged\n- add a focused hierarchical regression test and refresh affected routed snapshots\n\n## Game Boy Advance result\n\n- Pipeline 9 completes all 21 parent connections\n- relaxed DRC count improves from 34 to 30\n- four false same-net trace violations are removed\n- via count remains 144\n- final routed PCB snapshot is unchanged at 0.000 percent difference\n\nThis fixes DRC identity/reporting. It does not hide remaining physical DRCs or trade away connectivity.\n\n## Validation\n\n- bun run build\n- focused DRC and affected bug-report tests: 12 passed, 0 failed\n- full Game Boy Pipeline 9 repro: passed in 165 seconds, 0.000 percent snapshot difference\n- full suite before snapshot refresh: 570 passed, 60 skipped, with only the seven expected snapshot mismatches; all seven were inspected, refreshed, and pass in the focused rerun


Abse2001:
Closing this approach: it only corrected DRC net attribution and did not improve the Game Boy routing geometry. I am restarting from main to focus on Pipeline 9 preserved-subcircuit trace simplification without sacrificing connectivity.

PR #2221, opened 2026-08-24
Enforce per-connection minimum trace widths

seveibar:
## Summary
- add an explicit hard `minTraceWidth` constraint to SRJ connections while leaving `nominalTraceWidth` advisory
- make TraceWidthSolver test the effective hard minimum and fail when it cannot be routed, instead of silently emitting the board-level fallback width
- reject terminal tapering below an explicit hard minimum
- cover Pipeline9 propagation with feasible and impossible 0.18 mm minimum cases

## Validation
- `bun test tests/features/pipeline9-enforces-connection-min-trace-width.test.ts tests/features/tracewidthsolver/tracewidthsolver01.test.ts tests/features/tracewidthsolver/tracewidthsolver02.test.ts --timeout 9999999`
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`

## Follow-up
Pipeline9 path planning still plans around the board-level minimum width. This PR deliberately fails an impossible explicit per-connection minimum rather than undersizing it; making planning/rerouting width-aware is separate work.


seveibar:
Closing after an architecture review. The project intentionally retains usable autorouter output when a requested width cannot be met and reports the mismatch as a warning; this PR instead failed the whole solve late in TraceWidthSolver. It also rejected intentional terminal neck-down behavior, while path planning, connection split/merge logic, and repair do not carry the proposed hard constraint end-to-end. If hard per-connection guarantees are introduced later, they need a distinct SRJ/user contract, branch and terminal semantics, width-aware planning, and full-pipeline tests.

PR #2245, opened 2026-08-27
Reproduce preloaded via transition-coordinate drift

ShiboSoftwareDev:
## Summary

- adds a minimal preloaded-via merge reproduction
- models a serialized layer transition displaced by 0.0004 mm from its via
- snapshots the exact-coordinate lookup failure before applying any tolerance fix

This is the repro half of a two-PR stack.

## Validation

- focused Bun regression test passes
- SVG snapshot generated by graphics-debug


ShiboSoftwareDev:
Closing because this synthetic coordinate-drift state could not be reproduced from the AM62L board. The captured Pipeline9 inputs contain no transition-layer lookup failure, and the normal preloaded-trace conversion creates each via and its z-transition endpoints from the same x/y coordinates. We should reopen only with a real board-derived input that reaches this failure.

PR #85, opened 2025-03-27
Add Welcome page that allows uploading simple route json

seveibar:

PR #2246, opened 2026-08-27
Reproduce SII9022 HDMI routing failure without fanout identity

ShiboSoftwareDev:
Runs the pre-fix fanout on the real SII9022ACNU → TPD12S016PWR → HDMI-001S path. Pipeline 9 fails before producing its 16 HDMI route segments.


ShiboSoftwareDev:
This fixture was useful for locating the failure, but it contains invalid upstream fanout output with missing stable trace identities. The real root repro and fix are now tscircuit/fanout-solver#84 and #85, so this autorouter repro should not be merged.

ShiboSoftwareDev:
Superseded by tscircuit/core#3509 and tscircuit/core#3510. The real integrated repro and fix belong in core, where the invalid bus constraint is generated.

Issue #173, opened 2025-06-26
Autorouting Bug Report: <board#753 />


techmannih:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=bac7d21c-1248-4c48-9e22-07ee2477f344

(any additional details? Maybe a screenshot?)

PR #2248, opened 2026-08-27
Honor Pipeline9 bus allowed layers per connection

ShiboSoftwareDev:
Stacked on #2243.

## Real board failure

The regression is extracted from the AM62L carrier HDMI path and keeps the real `SII9022ACNU -> TPD12S016PWR -> HDMI_001S` three-point connection. `TMDS_D2_N` belongs to a bus whose `allowedLayers` is `["top"]`, but Pipeline 9 previously emitted part of that connection on `inner1`.

The SVG snapshot in #2243 shows the violating inner-layer segment. This PR updates that same snapshot to show the complete connection on `top`.

## Fix

- derives allowed z-layers for every final MST/point-pair connection from its logical bus member
- carries that constraint on each port point through tiny-hypergraph pathing and high-density routing
- uses a node-wide layer restriction only when every connection in the node has the same layer set
- keeps mixed per-connection constraints separate instead of applying the first port's layers to the whole node
- rejects forbidden tiny-hypergraph port transitions before cost calculation or candidate queueing
- validates the routed HD output against the declared bus layers before returning it

The pre-queue traversal hook is implemented by the standalone, `main`-based tscircuit/tiny-hypergraph#173. The dependency pin is a temporary compatibility commit built from current tiny-hypergraph `main`, #173, and the two unmerged solver APIs already consumed by the #2243 base (trace-density options and preferred-route preservation). It excludes the obsolete repro fixture and branch ancestry. This does not add a Pipeline 9 stage and does not add an `__allowedZ` field to SRJ connections.

## Validation

- real AM62L D2-pair regression passes and its SVG snapshot changes from the violation to top-only copper
- ten related Pipeline 9, preloaded-route, and tiny-hypergraph tests pass (11 focused tests total)
- TypeScript typecheck passes
- production build passes
- the compatibility branch's 19 focused tiny-hypergraph tests pass
- tiny-hypergraph's focused transition-filter regression proves rejected neighbors are neither costed nor queued

## Known follow-up

The reduced real pair routes correctly. The full eight-member HDMI phase now fails loudly when its current topology requires a single-layer crossing, rather than silently escaping the declared layers. That topology problem is separate from enforcing `allowedLayers`.


seveibar (commented):
this doesn't make sense- one connection limits the entire node to particular layers?

also why is __allowedZ being made part of the API?

You're hacking on bus routing to Pipeline9 without a plan

seveibar (changes requested):
meant to RC

ShiboSoftwareDev:
Rewritten from scratch on the real AM62L repro in #2243. The previous first-port/node-wide constraint bug and SRJ `__allowedZ` approach are gone; constraints now remain per connection, with node-wide restriction only when every connection agrees. The same real-board SVG snapshot is updated to show the fix. Ready for re-review once CI finishes.

lib/solvers/PortPointPathingSolver/tinyhypergraph/SelectiveReripTinyHyperGraphSolverWithStableInitialAssignments.ts
```diff
@@ -8,6 +13,46 @@ import { applyInitialAssignments } from "tiny-hypergraph/lib/initialAssignments"
...
+      ? this.allowedZByConnectionName[connectionName]
+      : undefined
+    if (allowedZ) {
+      const currentPortZ = this.topology.portZ[currentCandidate.portId]
+      const neighborPortZ = this.topology.portZ[neighborPortId]
+      if (
+        !allowedZ.includes(currentPortZ) ||
+        !allowedZ.includes(neighborPortZ)
+      ) {
+        return Number.POSITIVE_INFINITY
```
seveibar: this is a performance issue, you have to disable z traversal (it must never be queued if it is not allowed)

never use infinite or high costs to discourage, it is a huge performance issue!

seveibar: you should probably be modifying tiny-hypergraph etc.

ShiboSoftwareDev: Fixed in f91206c6. The autorouter no longer overrides computeG or returns Infinity for layer restrictions. tiny-hypergraph #173 adds a symmetric isPortTransitionAllowed hook and checks it before normal/outside-in candidate queueing, frontier joins, and selective relaxed expansion; this PR pins commit 979ec9a and overrides that hook per connection. The real AM62L snapshot is unchanged and passes, along with nine related Pipeline9/tiny-hypergraph tests. Upstream: https://github.com/tscircuit/tiny-hypergraph/pull/173

PR #2249, opened 2026-08-27
Fix Pipeline 9 multipoint bus length matching

ShiboSoftwareDev:
Fixes Pipeline 9 length matching for multipoint buses by grouping MST child routes under their root connection and using the shared length-matching postprocessor. The AM62L HDMI clock-pair repro in #2244 improves from 4.10 mm skew to within its 0.50 mm limit.


seveibar (changes requested):
no repro, ai

seveibar (changes requested):
no repro, ai

ShiboSoftwareDev:
Rewritten on the real AM62L clock-pair repro in #2244. The old implementation actually failed this fixture because each three-point bus member becomes two MST routes. The new version groups them by root connection, updates the same SVG snapshot, and keeps matching inside the existing post-processing stage. Ready for re-review once CI finishes.

lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/autorouting-pipeline-solver9-preloaded-trace-graph.ts
```diff
@@ -265,7 +265,7 @@ export class AutoroutingPipelineSolver9_PreloadedTraceGraph extends BaseSolver {
   deadEndSolver?: DeadEndSolver
   traceSimplificationSolver?: TraceSimplificationSolver
   mutatedPreloadedTraceSimplificationSolver?: TraceSimplificationSolver
-  lengthMatchingPostProcessingSolver?: DifferentialPairPostProcessingSolver
+  lengthMatchingPostProcessingSolver?: Pipeline9LengthMatchingPostProcessingSolver
```
seveibar: why is it different than Pipeline7's?

lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/autorouting-pipeline-solver9-preloaded-trace-graph.ts
```diff
@@ -1299,7 +1302,10 @@ export class AutoroutingPipelineSolver9_PreloadedTraceGraph extends BaseSolver {
 
   _getOutputHdRoutes(): HighDensityRoute[] {
     if (
-      (this.originalSrj.differentialPairs?.length ?? 0) > 0 &&
+      ((this.originalSrj.differentialPairs?.length ?? 0) > 0 ||
+        (this.originalSrj.buses ?? []).some(
+          (bus) => bus.maxLengthSkew !== undefined,
+        )) &&
```
seveibar: use `const hasLengthMatchPairs = ...`

seveibar (commented):
it doesn't need to be a pipeline9 specific solver, otherwise looks good

lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/autorouting-pipeline-solver9-preloaded-trace-graph.ts
ShiboSoftwareDev: Moved the wrapper to lib/solvers/length-matching-post-processing-solver.ts and renamed it LengthMatchingPostProcessingSolver. Pipeline 7 still maps each differential member to its single final HD route; Pipeline 9 additionally supplies logical bus roots because each AM62L member is split into two MST routes. The solver itself is no longer Pipeline 9-specific.

lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/autorouting-pipeline-solver9-preloaded-trace-graph.ts
ShiboSoftwareDev: Done. The output selection now computes const hasLengthMatchPairs before the conditional.

seveibar (approved):

PR #2250, opened 2026-08-27
Preserve preloaded via transitions across coordinate drift

ShiboSoftwareDev:
Stacked on #2245.

## Summary

- matches serialized via and transition coordinates within a 0.001 mm tolerance
- preserves the actual transition-layer point when moving a same-net via
- updates cluster replacement to use the same tolerance consistently
- changes the snapshot to show the transition following the retained via

## Validation

- focused same-net via merger tests pass
- TypeScript typecheck passes
- production build passes
- SVG snapshot diff reviewed


ShiboSoftwareDev:
Closing with its unreproducible base PR #2245. Without a real input that creates coordinate drift, adding a global tolerance to SameNetViaMergerSolver is not justified and could mask a different upstream invariant violation.

PR #2251, opened 2026-08-27
Connect preloaded copper at route endpoints

ShiboSoftwareDev:
Preloaded wire endpoints now share connectivity with route targets at the same position and layer. The AM62L repro changes from unrouted to all eight HDMI nets routed.


seveibar (changes requested):
ai

ShiboSoftwareDev:
Rewritten on the real 8-connection/48-preload AM62L fallback repro in #2246; the synthetic test is removed. The change is localized to the existing portfolio and grow/shrink solvers: rejected A03 is skipped and the valid A01 candidate becomes the winner. The same real SVG snapshot now shows that state transition. Ready for re-review once CI finishes.

seveibar (commented):
identical visual diff? what's happening, code seems ok

seveibar (changes requested):
invalid input problem

ShiboSoftwareDev:
You're right—the previous change was solving the wrong problem, so I removed it. The real AM62L input shows TinyHypergraph producing terminals at non-PCB intermediate targets inside an existing capacity node; the replacement fix makes the existing subdivision stage cut nodes at those layer-aware targets before pathing, so every high-density node in this repro is border-to-border.\n\nA global assertion also exposed a separate existing source of interior points in createRegionalFallbackProblem when a preloaded route is fully contained in the fallback node. I left that out of this PR because it affects nine existing Pipeline 9 cases and needs its own focused repro/fix.

ShiboSoftwareDev:
Follow-up from full CI: the first replacement also split ordinary two-endpoint fanout connections, which pushed the Pipeline 10 SRJ29 pathing repro over its budget. I narrowed the cut to non-PCB points in multi-point connections—the AM62L case is eight 3-point nets, while SRJ29 fanout is 16 ordinary 2-point nets. The exact Pipeline 10 test now passes, the AM62L clock route remains at 0.00 mm skew, and the Pipeline 9 suite passes 54/54.

ShiboSoftwareDev:
Reworked based on the review. #2246 now shows the real U5/U6/J2 board, all 48 preloaded fanout traces, all eight HDMI connections, and all three invalid nodes with 32 interior endpoints. The fix no longer changes the mesh before pathing: it splits only declared multi-point bus regions at the high-density handoff and preserves their common allowed layers, so the SVG changes from three red invalid regions to four green border-to-border regions. The clock pair is top-only with zero vias, zero relaxed-DRC errors, and 0.00 mm skew; all 55 Pipeline 9 tests pass.

ShiboSoftwareDev:
Final scope correction: the fix now runs only for declared multipoint bus connections, so the existing Pipeline 10 SRJ29 two-point route and SVG remain unchanged. The real U5/U6/J2 AM62L repro still changes from three invalid regions to four border-to-border regions, and its clock pair stays top-only with zero vias; all CI checks pass.

ShiboSoftwareDev:
Corrected the visualization itself: the previous snapshot used the real SRJ but rendered the packages as obstacle rectangles. #2246 now uses the exact emitted QFN-72 U5, TSSOP-24 U6, and HDMI J2 pad/silkscreen/hole geometry; the fix diff keeps that PCB and its 48 fanout traces unchanged and only moves the 32 marked terminals from red node interiors onto four green node boundaries.

ShiboSoftwareDev:
The previous node-splitting approach was incorrect and has been removed. This revision fixes the lost net identity at preloaded copper endpoints: the same real 8-net/48-preload AM62L input changes from no route to 16 generated HDMI segments (64 total traces), and the snapshot now shows unrouted → routed.

ShiboSoftwareDev:
The root cause is upstream: fanout-solver drops the stable source trace identity. It is reproduced with the real U5/TMDS fanout in tscircuit/fanout-solver#84 and fixed in tscircuit/fanout-solver#85. Adding only those identities makes unchanged Pipeline 9 solve this exact 48-preload input and emit 16 routes, so this endpoint-coordinate workaround is not needed.

PR #2258, opened 2026-08-27
Reduce Bug94 DRC errors with shared trace corridor repair

imrishabh18:
## Summary

- pin `high-density-repair03` to `a9654302bdfdbc93222f1f66ce5d67ccd6ecd891` from tscircuit/high-density-repair03#84
- lower Bug94's expected DRC ceiling from 8 to 6
- update only the affected Bug94 SVG snapshot

## Result

The shared-route layer corridor removes both remaining collisions involving `source_trace_97_mst3` without introducing replacement DRC identities, reducing Bug94 from 8 to 6 errors.

## Validation

- `bun test tests/bugs/bugreport94-56fa2e.test.ts --timeout 9999999`
- `bun run build`

Depends on tscircuit/high-density-repair03#84.


imrishabh18:
/benchmark --same-machine --dataset srj18 --concurrency 1 --sample-timeout 480s

ShiboSoftwareDev:
why did srj18 start timing out, is that falkiness or a regression?

ShiboSoftwareDev (approved):
we can merge if there are no regressions

PR #2282, opened 2026-08-28
Route SII9022 HDMI path with pinned fanout identity fix

ShiboSoftwareDev:
Pins fanout-solver fix 63f5123 on the same real SII9022ACNU → TPD12S016PWR → HDMI-001S path. The SVG changes from unrouted to all eight HDMI nets routed with 16 new segments.


ShiboSoftwareDev:
Superseded by tscircuit/core#3509 and tscircuit/core#3510. The routed result now comes from correcting core's bus input, with no pinned fanout or autorouter change.

PR #2293, opened 2026-08-29
perf: reduce high-density portfolio overhead

AnasSarkiz:
## Summary

- replace the high-density portfolio's repeated linear best-candidate scan with a stable binary min-heap
- update the selected heap root in place instead of popping, allocating, looking up its order, and reinserting it after every supervisor step
- preserve the legacy selection contract: solved candidates first, then lowest fitness, then original insertion order
- cache repeated future-connection cost queries in a bounded 1,024-slot direct cache
- validate exact node coordinates on cache hits so rounded grid-key collisions cannot change route costs or output
- retain the winning candidate after completion so visualization and post-solve inspection preserve legacy behavior

## Why

The portfolio can supervise dozens of candidates. The heap removes its linear scan, while the in-place root update removes the remaining pop/reinsert overhead.

CPU profiling then showed that queue management was below 0.1% of a representative large-node run. The actual hotspot was repeated future-connection scoring: the helper consumed about 23% total CPU, scanned roughly 30 points per call, and about 64% of requests were reusable with a bounded direct cache. The cache also reuses the already-computed planar distance and evaluates the via penalty once per miss instead of once per cross-layer point.

The cache is intentionally bounded. An earlier unbounded prototype was rejected because it added about 70 MB RSS in a 5,000-step stress run.

## Correctness

- heap ordering remains solved-first, then lowest fitness, then stable insertion order
- failed roots are removed and changed roots are reheapified without changing candidate priority
- cache entries require matching node key and exact `x`, `y`, and `z`
- rounded-key collision coverage alternates between two coordinates sharing one key and verifies the correct point is recomputed each time
- route completion, relaxed DRC, DRC issue count, and via count were unchanged in the local Pipeline9 comparison

## Validation

- `bun run build`
- `bun run format:check`
- 19 focused portfolio, future-cost, large-node, Pipeline9, and grow/shrink tests passed
- representative large-node stress run: 2,094.4 ms to 1,848.7 ms (**11.7% faster**)
- bounded-cache memory in that stress run: RSS 273.4 MB to 277.7 MB; live heap 60.7 MB to 62.3 MB
- local sequential Pipeline9/SRJ18 subset (7 samples):

| Metric | Baseline | PR | Delta |
| --- | ---: | ---: | ---: |
| Completion | 100.0% | 100.0% | 0.0 pp |
| Relaxed DRC pass | 85.7% | 85.7% | 0.0 pp |
| P50 time | 51.1s | 49.4s | **3.3% faster** |
| P95 time | 90.2s | 88.1s | **2.3% faster** |
| Average vias | 155.14 | 155.14 | 0.0% |

## Exact-head benchmark

[Full same-machine Pipeline9/SRJ18 benchmark](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33262048382) ran all 16 scenarios sequentially on one 8-vCPU Blacksmith runner, comparing main `85d904d` with PR head `f1f9e1d`:

| Metric | Main | PR | Delta |
| --- | ---: | ---: | ---: |
| Completion | 75.0% | 75.0% | 0.0 pp |
| Relaxed DRC pass | 50.0% | 50.0% | 0.0 pp |
| DRC issues | 74 | 74 | 0 |
| Timeouts | 2 | 2 | 0 |
| P50 time | 102.1s | 104.0s | **1.9% slower** |
| P60 time | 129.1s | 128.2s | 0.8% faster |
| P70 time | 227.0s | 224.9s | 0.9% faster |
| P80 time | 270.5s | 271.0s | 0.2% slower |
| P95 time | 360.0s | 360.0s | 0.0% |
| Average vias | 187.50 | 187.50 | 0.0% |

Outcome changes: 0 improved, 0 regressed.

The stage-level artifact shows that the target high-density route stage did improve in aggregate: 735.9s to 717.4s (**2.5% faster / 18.5s saved**). The total Pipeline9 percentile result remains effectively neutral because unrelated stages and samples that never reach high density dominate the variance.

## Current conclusion

The optimization is measurable inside the target high-density stage and preserves routing outcomes, but the full exact-head benchmark does not show a meaningful end-to-end Pipeline9 speedup. This PR should remain a draft and should not be merged as a general performance improvement without a larger gain.

### DIFF
--- a/lib/solvers/HighDensitySolver/SingleHighDensityRouteSolver6_VertHorzLayer_FutureCost.ts
+++ b/lib/solvers/HighDensitySolver/SingleHighDensityRouteSolver6_VertHorzLayer_FutureCost.ts
@@ -2,6 +2,8 @@ import { distance, pointToSegmentDistance } from "@tscircuit/math-utils"
 import { SingleHighDensityRouteSolver } from "./SingleHighDensityRouteSolver"
 import { Node } from "lib/data-structures/SingleRouteCandidatePriorityQueue"
 
+const FUTURE_CONNECTION_POINT_CACHE_SIZE = 1024
+
 export class SingleHighDensityRouteSolver6_VertHorzLayer_FutureCost extends SingleHighDensityRouteSolver {
   FUTURE_CONNECTION_PROX_TRACE_PENALTY_FACTOR = 2
   FUTURE_CONNECTION_PROX_VIA_PENALTY_FACTOR = 1
@@ -10,8 +12,11 @@ export class SingleHighDensityRouteSolver6_VertHorzLayer_FutureCost extends Sing
   VIA_PENALTY_FACTOR_2 = 1
   FLIP_TRACE_ALIGNMENT_DIRECTION = false
   FUTURE_CONNECTION_VIA_TRACE_CLEARANCE = 0.1
-  futureConnectionPoints: Array<{ x: number; y: number; z: number }>
+  futureConnectionPoints: FutureConnectionPoint[]
   futureConnectionSegmentsCache: FutureConnectionSegment[] | null = null
+  private closestFuturePointCache: Array<
+    ClosestFuturePointCacheEntry | undefined
+  > = []
 
   constructor(
     opts: ConstructorParameters<typeof SingleHighDensityRouteSolver>[0],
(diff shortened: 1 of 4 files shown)


AnasSarkiz:
/benchmark --same-machine --pipeline 9 --dataset 18 --sample-numbers 1,3,5,9,10,11,16 --concurrency 1 --sample-timeout 15m

AnasSarkiz:
/benchmark --same-machine --pipeline 9 --dataset 18 --sample-numbers 1,3,5,9,10,11,16 --concurrency 8 --sample-timeout 15m

AnasSarkiz:
Closing this draft because the measured priority-queue change only reduced portfolio overhead by about 1–2% and does not address the native-size high-density failures that trigger growth. The focused root-cause work is now in tscircuit/high-density-a01#96 with production verification in #2319.

PR #2301, opened 2026-08-29
superseded: Pipeline 9 B02 coverage moved to #2298

seveibar:
Superseded by #2298.

B02 is now selected automatically for dense original-scale four-layer portfolio nodes, so Pipeline 9 needs no separate rollout. The focused Pipeline 9 integration test and its `toMatchSvgSnapshot` coverage moved into #2298; the two-layer SRJ23 and Game Boy baselines remain unchanged and were dropped from the B02 diff.


seveibar:
Superseded by #2298 now that B02 is an automatic original-scale candidate and the Pipeline 9 SVG coverage lives with the adapter.

PR #2315, opened 2026-08-30
fix: align single-transition boundary tolerance

AnasSarkiz:
## Summary

- establishes one shared internal `1e-6 mm` boundary-coordinate tolerance
- uses that policy consistently in point classification and boundary-angle calculation
- snaps an accepted near-boundary input to the exact nearest boundary before later geometry work
- updates the real SRJ18 fixture to prove the solver no longer throws and stays deterministic
- adds all-side and corner tolerance coverage plus a guard proving authoritative DRC still rejects real copper contact

## Native solver visual

| Before: inconsistent rejection | After: candidate generation reached |
| --- | --- |
| before | after |

The after snapshot is evidence that the floating-point inconsistency is removed, not that the generated candidate is DRC-valid. The extracted candidate still overlaps copper and must be rejected by downstream authoritative validation. This PR makes no SRJ18 performance or solve-rate claim.

## Validation

- focused reproduction, boundary, and authoritative-DRC tests: 3 pass / 0 fail
- typecheck passes
- build passes
- full suite was exercised; unrelated HTTP tests cannot bind ephemeral ports in the sandbox, and the same representative HTTP test passes outside it

Stacked on the reproduction PR so the before/after behavior and visual diff are directly reviewable.


seveibar (approved):
very nice

PR #2322, opened 2026-08-31
feat: report high-density growth attempts in benchmarks

seveibar:
## Summary

- count each high-density grow-and-retry transition once, at the transition in `GrowShrinkHighDensityIntraNodeSolver`
- share one typed counter across Pipeline9's ordinary and regional high-density solvers, so replacing a child solver does not require commit/reconstruction logic
- expose the count through a typed solver method instead of untyped `stats`; networked Pipeline9 explicitly reports it as unsupported because remote responses do not contain growth telemetry
- retain the latest observed count when a benchmark worker is interrupted, mark it partial, and render aggregate partial values as lower bounds (`≥N`)
- show the metric in standard and same-machine tables only when at least one report supports it; do not calculate a delta between lower bounds
- render paired results with the PR checkout's formatter, so a newly introduced metric can appear before `main` knows its schema; the redundant controller checkout is removed

`HD growth attempts` counts scale transitions, not unique nodes.

## Progress-path safety

The worker reads only the O(1) counter after the existing progress throttle accepts a heartbeat. Full TinyHypergraph routing metrics remain final-result-only. The regression test runs 1,000 solver steps with a 60-second throttle and verifies that only the initial and final counter snapshots are collected.

## Validation

- focused benchmark/growth tests: 7 passed, 46 assertions
- grow/shrink integration plus Pipeline9 and bug regression tests: 10 passed, 59 assertions
- networked Pipeline9 growth-telemetry contract test passes
- exact same-machine workflow-layout replay of run `33444481724` renders `Pipeline9 | HD growth attempts | n/a | ≥120 | n/a`
- workflow YAML parses successfully and contains no stale `same-machine-controller` reference
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`

## Benchmark

[Fresh default-timeout same-machine Pipeline9 SRJ18 run](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33444481724), comparing `main` `a35631a` with runtime-equivalent PR commit `4a0b8ba`:

- completion: 75.0% → 75.0%
- relaxed DRC pass: 50.0% → 50.0%
- timeouts: 2 → 2
- outcome changes: 0 improved, 0 regressed
- P50 +1.6%, P60 -0.8%, P70 -1.2%, P80 +0.1%, P90 -0.4%, P95 unchanged
- average vias unchanged

The PR report contains `{ value: 120, status: "partial" }`. Of that, 53 attempts are known complete and sample 6 contributed a last observed 67 before timing out during high-density routing. The honest aggregate is therefore **≥120**. Current `main` predates metric serialization, so the comparison row is:

| Solver | Metric | Main | PR | Delta |
| --- | --- | ---: | ---: | ---: |
| Pipeline9 | HD growth attempts | n/a | ≥120 | n/a |

The official comment for that run omitted the row because the default-branch workflow rendered it with `main`'s older formatter. This PR now renders from the already-checked-out and installed PR revision; an exact raw-report replay verifies the row above. Comment-triggered workflows themselves run from the default branch, so that workflow-selection change becomes observable in normal `/benchmark --same-machine` comments once this PR reaches `main`.

Because `main` does not report the metric yet, this prerequisite PR cannot calculate a growth delta. Subsequent A11 comparisons can report both sides with the same schema.


seveibar:
## Growth-row preview (corrected)

The default-timeout same-machine run at `df4a990` kept outcomes unchanged: completion remained 75.0%, timeouts remained 2, and there were no outcome changes. Timing was mixed, including a P60 increase of 5.5%. [Official result](https://github.com/tscircuit/tscircuit-autorouter/pull/2322#issuecomment-5484265194).

The PR artifact observed 120 growth attempts, but sample 6 timed out during high-density routing after its latest heartbeat had observed 67 attempts. The total is therefore a lower bound, not an exact count. The updated formatter renders:

| Solver | Metric | Main | PR | Delta |
| --- | --- | ---: | ---: | ---: |
| Pipeline9 | HD growth attempts | n/a | ≥120 | n/a |

`main` correctly shows `n/a` because it predates metric serialization. The source-counter refactor at `18090f23` now preserves partial/exact status explicitly and suppresses deltas between lower bounds.

seveibar:
## Fresh growth-count result

[Run 33444481724](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33444481724) kept outcomes unchanged (75.0% completion, 2 timeouts, 0 improved / 0 regressed). The PR artifact reports `{ value: 120, status: "partial" }`: 53 attempts are known complete, plus at least 67 observed before sample 6 timed out inside high-density routing. The correct total is therefore **≥120**.

| Solver | Metric | Main | PR | Delta |
| --- | --- | ---: | ---: | ---: |
| Pipeline9 | HD growth attempts | n/a | ≥120 | n/a |

The official bot comment omitted this row because the default-branch workflow rendered with `main`'s older formatter. Head `b700331` now renders from the already-installed PR checkout and removes the redundant controller checkout. Replaying the run's exact raw reports produces the row above. Normal comment-triggered runs will exercise that workflow change after it reaches `main`.

PR #2329, opened 2026-09-01
feat: add bounded A11 and A12 native routing to Pipeline9

AnasSarkiz:
## Summary
- enable upstream A11 and A12 as Pipeline9 candidates only at native node size
- preserve established solver results, then try A11, then the more expensive A12
- keep both fine-grid candidates lazy and bound them to 5,000 and 15,000 iterations respectively
- preserve the existing grown-node portfolio and reject invalid geometry or disconnected port pairs
- use the single canonical high-density dependency

## Root cause
Some difficult high-density nodes only converged after grow/shrink because the production Pipeline9 portfolio did not include the specialized native-grid solvers. A11 covers fine-grid history-aware searches, while A12 adds a mixed-resolution search for cases where a uniform fine grid is unnecessarily large.

A12 previously caused a timeout when it could run too early and search for too long. Here it starts after established candidates and A11, is limited to the native-size attempt, and stops after 15,000 iterations. The exact SRJ18 sample 12 regression now completes DRC-clean without increasing any timeout.

## Validation
- package build, TypeScript typecheck, formatting, and diff checks
- focused A11 and A12 integration, priority, lazy setup, iteration-bound, native-only, geometry, and connectivity tests
- exact Pipeline9 SRJ18 sample 12 replay: solved with zero relaxed-DRC errors
- full Game Boy Advance routing regression with the original SVG snapshot
- historical bugreport87 and bugreport88 routing checks
- grow/shrink and invalid-output regression coverage

### DIFF
--- a/lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/Pipeline9HighDensitySolver.ts
+++ b/lib/autorouter-pipelines/AutoroutingPipeline9_PreloadedTraceGraph/Pipeline9HighDensitySolver.ts
@@ -362,6 +362,7 @@ export const createPipeline9RegularNodeSolver = ({
     layerCount,
     useGrowShrinkHighDensityIntraNodeSolver: true,
     useHighDensitySolverA11: true,
+    useHighDensitySolverA12: true,
     preserveTerminalPcbPortIds: false,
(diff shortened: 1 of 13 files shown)


AnasSarkiz:
Closing this draft to split the A11 and A12 integrations into focused changes.

PR #2345, opened 2026-09-02
Add real vertical DDR3 pair reproduction

ShiboSoftwareDev:
## Summary

- add a regression fixture from the published BeagleBone Black U12-to-U5 DDR3 connection map
- rigidly rotate the complete two-package layout to exercise Pipeline 10 with a vertical BGA pair
- preserve all 16 audited nets, both buses and differential pairs, and all 96 plus 324 package balls
- snapshot the current rejection with the complete unrouted ratsnest visible

This is the reproduction/base PR for a stacked Pipeline 10 geometry fix. The follow-up PR uses this exact fixture and changes the SVG from the rejected ratsnest to successful fanout on both packages.

## Verification

- bun test tests/features/pipeline10-real-ddr-vertical-pair.test.ts
- bun run build
- bun run format:check


ShiboSoftwareDev:
Stacked root fix and SVG before/after proof: https://github.com/ShiboSoftwareDev/tscircuit-autorouter/pull/3

That review is based directly on this reproduction commit, so its Files changed view contains only the Pipeline 10 axis fix plus the test and SVG transition from rejected ratsnest to successful two-BGA fanout.

ShiboSoftwareDev:
Superseded by the upstream-owned E2E stack #2346 and #2347. The replacement keeps the test source unchanged in the fix PR.

PR #2346, opened 2026-09-02
Add real DDR pair end-to-end coverage

ShiboSoftwareDev:
## Summary

- add an end-to-end Pipeline 10 fixture using the published BeagleBone Black U12-to-U5 DDR3 map
- preserve all 16 audited nets, buses, differential pairs, and the complete 96-ball plus 324-ball package populations
- place the real packages in a vertical-dominant diagonal arrangement while retaining a valid horizontal corridor
- verify both BGA fanouts and the handoff into global autorouting, with a committed SVG

This is the base of a two-PR stack. The test passes before and after the production fix and is intentionally unchanged in the fix PR. Its baseline SVG records Pipeline 10 choosing horizontal fanout for this placement.

## Verification

- bun test --timeout 9999999 tests/e2e-pipeline10-real-ddr-pair.test.ts
- bunx tsc --noEmit

PR #2347, opened 2026-09-02
Make Pipeline 10 BGA pair fanout axis-aware

ShiboSoftwareDev:
## Stack

Depends on #2346. Both branches now live directly in tscircuit/tscircuit-autorouter.

## Root fix

Pipeline 10 previously sorted every BGA pair left-to-right, measured only horizontal clearance, and derived horizontal inward and outward fanout directions. This change selects the axis from the actual physical package-to-package gap, sorts and measures on that axis, and derives inward, outward, and perpendicular escape directions consistently for both fanout stages.

## Review shape

The E2E test source is unchanged from #2346. This PR modifies exactly:

- Pipeline 10 production solver geometry
- the E2E SVG snapshot produced by the unchanged test

The SVG diff visibly changes the same real two-package DDR circuit from horizontal-edge fanout to fanout based on its larger vertical corridor.

## Verification

- bun test --timeout 9999999 tests/e2e-pipeline10-real-ddr-pair.test.ts
- bunx tsc --noEmit
- bun run build


ShiboSoftwareDev:
Superseded by tscircuit/core#3596. The existing Pipeline 10 solver already provides the two-BGA fanout behavior; the missing integration is in core, so no autorouter patch is needed.

PR #2417, opened 2026-09-05
Fix DRC conversion and retain 15 independently verified SRJ33 failures

seveibar:
SRJ33 sample011 reported disconnected endpoints even though the trace ended inside its pads. DRC conversion dropped legacy metadata and deduplicated different rectangles belonging to the same pad. This fixes those errors and pins SRJ33 to 15 inputs whose 324 reported Pipeline 9 errors all have independently confirmed geometric witnesses.

The converter preserves distinct copper fragments and legacy oval shapes, respects explicit pad ownership, and retains pre-existing trace connectivity. Logical net membership comes from requested terminals and original identities, so a routed endpoint on the wrong net cannot merge the nets during conversion.

The [fresh Blacksmith audit](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33986579884) rerouted all 37 previous SRJ33 inputs with the corrected converter. Sixteen pass DRC, including sample011. Six more are excluded because not every error could be independently authenticated: legacy copper geometry or terminal evidence remains ambiguous. The 15 retained cases have 324 verified error records. Their source inputs, complete routing results, exact error objects, and all inclusion/exclusion decisions are preserved in [the dataset](https://github.com/tscircuit/dataset-srj33-drc-failures/tree/026a78cb005ab33dde24f2db8fefbfd8d8efa614).

An independent Shapely verifier measures clearances on raw SRJ/routed geometry, checks shared layers and declared net identities, and uses conservative inner copper where legacy geometry is ambiguous. Every reported error must pass; cases with unresolved errors are excluded even when other errors are valid. Negative controls reject separated copper, same-net contact, and different-layer crossings. These are the benchmark's relaxed 0.1 mm clearance checks on supplied SRJs.

Validation on `dad4f44`: the full format check, type check, builds, and all nine test shards pass. The [fresh default Pipeline 9 benchmark](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33995845730) routes all 85 cases with zero DRC issues or timeouts, matching main's completion and DRC outcomes. The [fresh SRJ33 benchmark](https://github.com/tscircuit/tscircuit-autorouter/actions/runs/33995846341) routes all 15 retained cases with 324 DRC issues and zero timeouts; every case's error count matches the authenticated dataset. These runs execute `./benchmark.sh --pipeline 9` and `./benchmark.sh --pipeline 9 --dataset srj33 --concurrency 8 --sample-timeout 1800s` on Blacksmith.

Dataset CI also passes membership/hash checks, exact DRC/conversion reproduction for all 15 retained cases, and independent verification of all 324 errors. The readability refactor was checked against all 37 archived samples: converted Circuit JSON and all 348 archived DRC errors were unchanged.


seveibar:
/benchmark --pipeline 9 --dataset srj33 --concurrency 8 --sample-timeout 1800s

PR #2419, opened 2026-09-05
Fix Pipeline9 fanout wrinkles at the simplification handoff

seveibar:
## Summary

Fix the jagged phase-two pad approaches in core's `breakout-bga36-decoupling-caps-autorouting-srj` snapshot with a narrow metadata correction at Pipeline9's existing simplification handoff.

Preloaded HD sections currently carry a canonical net key in `rootConnectionName`. The simplifier resolves that field as a connection ID through `connMap`, so it can treat a route's own phase-one fanout as foreign copper. Generated net keys may also collide with IDs inherited from another connectivity map.

Preserve `trace.connection_name` as the root on the copies handed to simplification. Keep the shared converter unchanged: B01 and other routing stages intentionally use canonical roots.

- Five added production lines; no flags, API changes, dependencies, new geometry pass, or relaxed clearances.
- Add the exact BGA phase-two input and an SVG regression with the benchmark DRC overlay. All eight pad approaches are straight or single-bend; all eight preloaded traces remain unchanged; DRC errors are zero.
- Add a focused real-handoff test for same-net aliases and foreign copper, also asserting the shared converter retains canonical roots.
- Refresh the existing SRJ23 visual test snapshots affected by correctly recognizing preloaded copper.

## Validation

- Both new regression tests fail with the production fix removed and pass with it restored. The BGA reproduces six-point wrinkles on the base code.
- 21 focused tests across 17 files pass, including preloaded-route immutability, through-obstacle reconstruction, repair02, SRJ23 visuals, and differential-pair pad clearance.
- `bunx tsc --noEmit` passes.
- `bun run build` passes.
- Visually inspected the new BGA SVG and the changed SRJ23 circuit010 snapshot.

No full benchmark run has been performed for this PR.


seveibar:
Updated the failing tests in f5a2680a:

- Refreshed the three affected SVG snapshots after inspecting their before/after renders: QFP16, HDMI clock pair, and Game Boy.
- Corrected the three formatting differences reported by CI.
- Updated the QFP16 caption and added an assertion for its reduction from three vias to one.
- No snapshot tolerances or existing routing assertions were relaxed.

Local validation: 24 focused tests pass on Bun 1.3.8, including non-update reruns of all three previously failing tests. Typecheck passes. The new CI format, typecheck, and build checks have passed; the longer test shards are still running.

**Do not treat the refreshed Game Boy snapshot as a clean routing improvement.** An additional focused before/after relaxed-DRC check found a regression that needs attention before merging:

| Fixture | Relaxed DRC errors before → after | Routed vias before → after |
| --- | --- | --- |
| QFP16 | 0 → 0 | 3 → 1 |
| Game Boy | 35 → 37 | 144 → 146 |

The Game Boy comparison restores the previous simplification handoff on a separate solver instance, with all other code and inputs unchanged. Five error pairs are introduced and three resolved. The new errors include cross-net overlaps involving `source_net_7_mst3_0` and `source_trace_65_0`, plus trace/via clearance violations; connMap confirms these are foreign nets, not just ownership-label aliases. This is not a full benchmark run.

PR #2430, opened 2026-09-06
Match Pipeline9 final DRC validation with Core

Abse2001:
## Summary

- make Core's complete routing DRC the final acceptance and rollback authority for Pipeline 9
- align the reusable Repair03 geometry evaluator with Core's declared trace, via-to-pad, and drill-hole spacing rules
- reject geometrically invalid candidates through the indexed evaluator before paying for a complete Core Circuit JSON check
- preserve one Repair03 portfolio; there is no second repair pass and no union of conflicting DRC objectives

## Why

Pipeline 9 could finish according to its optimized DRC objective while Core still rejected the resulting board. Evaluating every mutation through Core fixed correctness but caused serious timeouts because it rebuilt and rechecked the full circuit for each candidate.

The updated design is a Core-gated cascade. The indexed evaluator cheaply rejects known-bad geometry, while candidates that pass it must also pass the full Core evaluator. Core also guards accepted Repair03 and post-precision outputs from DRC regressions. This keeps final behavior authoritative without adding a Game Boy- or dataset-specific path.

Repair03 parity changes are stacked in tscircuit/high-density-repair03#102.

## Validation

- TypeScript typecheck passes
- focused Core parity, reference-gate, cache/connectivity, maximum-via, and Pipeline 9 tests pass
- bugreport94 routes successfully and its affected macOS snapshot is updated
- full CI and same-machine benchmarks run remotely


Abse2001:
/benchmark --pipeline 9 --dataset 23 --sample 52 --same-machine

PR #2446, opened 2026-09-07
Route traces within per-trace via limits

Abse2001:
## Summary

- preserve each source trace maxViaCount through connection merging, point-pair splitting, and off-board point substitution
- reject or retry global and high-density routing candidates that exceed the selected source connection via budget
- validate completed paths per original connection so adjacent branches of a merged net do not consume each other’s budget
- add focused and real-board regressions showing zero-via traces remain on one layer

This refreshes and supersedes the auto-closed #2175/#2184 stack on current main. The conflict resolution preserves current preloaded-trace optimization behavior and disables the incompatible optimizer only for inputs carrying max-via constraints.

Crystal traces keep their zero-via constraint, and any routed trace that violates its maxViaCount remains an error. Impedance matching is intentionally out of scope.

## Validation

- 7 focused max-via and real-board tests passed
- bunx tsc --noEmit
- bun run build
- git diff --check


Abse2001:
Closing after design clarification. Per-connection maxViaCount is a difficult global-routing constraint and is not currently part of the released general autorouter input. For crystal traces, the preferred solution is to pre-route the trace explicitly with pcbPath and retain the existing max_via_count=0 error as a backstop. General router support can be reconsidered separately if broader use cases justify it.

PR #2473, opened 2026-09-08
Reuse port-point output for Pipeline9 node Pf calculations

AnasSarkiz:
## Summary

Pipeline9 already constructs the port-point output before preparing high-density routing. Reuse that output for each node's Pf calculation instead of rebuilding the complete output for every node.

- Two production files changed: 10 additions and 3 deletions.
- Keep the existing one-argument API, first-match lookup, missing-node/region behavior, and calculation order.
- No dataset-specific conditions, new flags, attempt thresholds, persistent cache, fallback behavior, or dependency changes.
- Only the two production files are included; no test files or benchmark tooling.

## Validation

- Repository formatter passed on both changed source files.
- Tests and benchmarks were not rerun for the test-removal update.
- Diff whitespace check passed.
- Branch updated to main at `207c4191`.
- The earlier current-main local typecheck was blocked by the missing fixture dependency `@tscircuit/fanout-for-pipeline9-fixtures`. CI must verify the current head.

A prior same-machine SRJ18 sample 6 pair on `f55494c7` measured 217.342s to 200.469s end-to-end (7.8% less time), with identical output, ordered DRC findings, 97 existing DRC issues, and 300 vias. This is one sample pair, not a full-suite or current-main performance claim. Benchmark tooling and artifacts are not included.


seveibar (approved):

seveibar:
@AnasSarkiz you definitely get credit for this fix (100%) but my AI rewrote it and I think it's a bit more understandable https://github.com/tscircuit/tscircuit-autorouter/pull/2475

rather than iterate on naming i'm going to merge that PR but this fix credit should go to you

PR #2506, opened 2026-09-09
Fix trace widening that creates copper clearance DRCs

seveibar:
Trace widening could turn valid clearances into DRC violations. On [bug report f6cc2bb3](https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=f6cc2bb3-c4bd-4f2f-ac41-5b11470e4234), the width stage increased relaxed-reference DRCs from 49 to 110. This change keeps that stage at 49 and lets the downstream pipeline finish with zero DRCs.

The width solver now queries the full required pad-clearance envelope, measures via copper using the via radius, and reindexes each completed route so later width choices respect already-widened copper. It validates continuous segments before accepting a width, covering short routes and gaps between cursor samples. Rotated pads are checked in their local coordinates. Existing nominal/midpoint/minimum width selection and terminal tapering remain in place.

Validation:

- Four new regression tests cover the pad query envelope, via radius, previously widened routes, and sub-step-length routes. All four fail on unchanged main and pass with this fix. The adjacent-route test also verifies nominal width is retained where it fits.
- Focused width/jumper and Pipeline 9 copper-width tests pass. Reviewed and updated the width snapshot, eleven changed Linux/full-board snapshots from CI artifacts, and four corresponding macOS snapshots.
- Updated the Pipeline 7 DRC regression to permit an unchanged intermediate baseline and a different repair branch after width selection changes. It still requires non-increasing repair counts and zero final DRC errors.
- `bun run build` passes, including declaration generation.
- Replayed the report using Pipeline 9, effort 1, an initially empty per-process cache, and the same dependencies as main `826b99f5` (v0.0.896). Counts use `evaluateRelaxedDrc`, including original board copper:

| Stage | Main | This PR |
| --- | ---: | ---: |
| Before width selection | 49 | 49 |
| Width selection | 110 | 49 |
| Global repair | 100 | 44 |
| Joint repair | 6 | 0 |
| Final output | 5 | 0 |

This is a focused report replay, not a dataset-wide benchmark. Existing violations entering the width stage still require downstream repair.


seveibar:
@Abse2001 merged. Should fix gameboy issues 🤞

PR #2536, opened 2026-09-10
use copper-pour-aware rectdiff regions for rv1106 pathing

mohan-bee:
### Motivation
- Address the node-construction review concern using [rectdiff #148](https://github.com/tscircuit/rectdiff/pull/148), stacked on the exact RV1106 [rectdiff repro #147](https://github.com/tscircuit/rectdiff/pull/147).

### Before
- The [autorouter repro #2534](https://github.com/tscircuit/tscircuit-autorouter/pull/2534) uses separate top and inner2/bottom free meshes across the inner1 copper pour.

### After
- Pin rectdiff to `30ebd720b25c0c4264de2c1ee58200f18b25beb4`, refresh the same full-board pathing snapshot, and retain the 30-second pathing assertion. The focused local test passed in 15.4 seconds, including snapshot checking; this is not a same-machine benchmark claim.
- Remove the custom autorouter topology-merging changes. Rectdiff now splits eligible overlaps at node construction, preserves remainders, and checks copper-pour coverage, solid obstacles and minimum via dimensions. Tiny Hypergraph, effort and solver iteration limits remain unchanged. These are path-planning snapshots, not completed copper routing or DRC clearance.

| Before | After |
| --- | --- |
| RV1106 pathing before | RV1106 pathing after |


seveibar (commented):
Based on the video i dont think this is conceptually valid

there is no way to merge overlapping free nodes that dont have the exact same bounds without shrinking to only the overlap area and stitching across the remaining, or forming new nodes where the overlapping area becomes a multilayer and subtracted from the others- in which case this should be part of whatever solver does node construction

mohan-bee:
working

PR #2546, opened 2026-09-11
Add exact T113 supervisor PCB SVG repro

ShiboSoftwareDev:
Pipeline 9 fails on the exact 41-component T113-S3 supervisor board after all three fanout solvers complete. The reproduction contains the complete captured Circuit JSON and routing input: 41 source and PCB components, 222 PCB ports and obstacles, 72 connection endpoints, and 40 native fanout traces. No geometry, connection, route, or via was reduced or manually replaced.

The test confirms that the captured PCB has no materialized traces or vias, verifies the original B01 failure, and renders the full Circuit JSON through `convertCircuitJsonToPcbSvg`. The committed `.snap.svg` is therefore the actual unrouted PCB. The stacked fix in #2549 changes this same snapshot path to the 94-trace routed result, which gives reviewers a direct PCB SVG image diff.

Validation: `bun test tests/bugs/t113-linux-supervisor-boundary-clearance.test.ts --timeout 9999999`; `bunx tsc --noEmit`.


ShiboSoftwareDev:
Root fix stacked in #2549. It keeps Pipeline 9 boundary ports far enough from mesh-edge corners to satisfy B01’s preloaded-trace clearance, and the exact T113 fixture now routes successfully.

PR #2553, opened 2026-09-12
Guarantee no different-net same-layer shorts in routed output (#1507)

### LINKED ISSUE #1507: Same-layer crossings can survive to output as net shorts (crossings modeled as via-resolvable, not guaranteed resolved)

maci0:
Replaces the stale-bot-closed #1513.

The capacity pipeline models same-layer crossings as via-resolvable *cost* (`calculateCrossingProbabilityOfFailure`), not a hard constraint. When HD/repair cannot actually resolve one, it survives as a different-net same-layer overlap, a real short. Nothing in the output path checked for that.

This adds `guaranteeNoSameLayerShorts` and runs it at the public output of Pipeline 7 (`AutoroutingPipelineSolver`) and Pipeline 9 (the current default in `@tscircuit/core`). A violating segment is truncated to a ratsnest instead of shipping a short.

The original dense-crossbar pipeline fixture from #1513 now throws in `CrossingViaReductionSolver` (`layer transition without a via`), so the tests cover the safety net directly: a crossing pair is a DRC overlap before the pass and not after.

Follow-up, not in this PR: resolve the crossing with an inserted via instead of truncating.

Fixes #1507


maci0:
Updated the two CI snapshot failures:

- \`bugreport44-0ec411\`: last-step SVG drifted by a handful of pixels (same topology).
- \`bugreport77-07f6a7\`: same-layer crossing segments are gone. That is the guarantee (open ratsnest instead of a short). The DRC assertions on that test still pass (\`< 94\` errors, no via errors).

maci0:
Stale bot closed this one unreviewed and it cannot be reopened (422 on both GraphQL and REST), so here is the fresh version against current main: #2647. Same change, rebased onto `8fae1e7c`, with the pre-existing `tests/bugs` main failures documented in the PR body.

PR #2556, opened 2026-09-12
Fix Pipeline9 through-via policy

Abse2001:
## Summary
- makes Pipeline9 treat vias as through-board by default when blind and buried vias are disabled
- preserves endpoint-only spans when the SRJ explicitly enables blind and buried vias
- updates high-density-repair03 to the Repair03 through-via DRC fix at 97ef741

## Reproduction
The focused repro was opened first in #2555. This PR includes that test as a passing regression after the fix.

## Validation
- bunx tsc --noEmit
- focused Pipeline9 policy, fixed-copper fallback, regional fallback, and preloaded-trace tests


Abse2001:
Superseded by #2557, which carries the same fix on an upstream branch so the repository snapshot updater can commit directly.

PR #2561, opened 2026-09-12
Connect same-region connections from degenerate HD routes

maci0:
Replaces stale-bot-closed #1530.

When both endpoints of a connection sit in one capacity node, terminal ports are dropped from the HD node graph and the solver can return a single-point route. The net is then left open. Fall back to a direct segment between the connection endpoints (with a via if the layers differ).


maci0:
Stale bot closed this one unreviewed, and it cannot be reopened (GitHub returns 422 on both GraphQL and REST), so I opened a fresh PR against current main: #2646. Same change, rebased onto `8fae1e7c`.

Issue #131, opened 2025-06-05
Autorouting Bug Report: <group#680 name=".U1" />


ShiboSoftwareDev:
Bug Report Link: https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=7ef5b44e-4d6d-4d1a-aa59-688de0557822

(any additional details? Maybe a screenshot?)

ricohageman:
@seveibar this board routes completely, can you close the issues?
<img width="452" height="596" alt="Image" src="https://github.com/user-attachments/assets/d04c672d-6ac6-4040-a679-cf3a099e1171" />

PR #2577, opened 2026-09-14
Honor physical through-via spans in Pipeline 9

Abse2001:
On a four-layer board with blind and buried vias disabled, a top-to-inner-layer via still has copper on the bottom layer. Pipeline 9 and its repair dependencies used only the logical endpoint span in several physical-clearance paths, so a bottom-layer route could cross that through via.

Use the existing `allowBlindAndBuriedVias` board policy consistently for fixed-route geometry, B01 obstacle selection, regional conflicts, Repair03, Repair04, and reference Circuit JSON export. Through vias occupy the complete board stack by default; the existing inclusive endpoint span remains available when blind and buried vias are enabled. Logical route endpoints remain unchanged.

Pin [Repair03 #131](https://github.com/tscircuit/high-density-repair03/pull/131) and [Repair04 #15](https://github.com/tscircuit/repair04/pull/15), which apply the same physical rule inside their existing via-clearance and DRC paths. No new routing mode or feature flag is introduced.

Regression coverage reproduces [#2555](https://github.com/tscircuit/tscircuit-autorouter/pull/2555), covers preloaded vias, multiple layer counts, reversed and coincident endpoints, explicit blind/buried opt-in, cache reuse, bottom-layer B01 routing, and reference/indexed DRC agreement. Existing fixtures that intentionally model partial vias now declare the existing opt-in.

Validation on current `main` (`1e4cfac1`): the three focused via regressions pass with 213 assertions, the existing pedometer graph remains green with 5,532 assertions, and typecheck and build pass. The complete GitHub test matrix is authoritative for the full suite. Benchmarks are run only through PR commands.


seveibar (commented):
You shouldnt be leaking Pipeline9 semantics here

seveibar:
Ie do not mention pipeline9

seveibar:
Also do nit conditionally gate on pipeline9 if thats whats happening

PR #2580, opened 2026-09-14
Add T113 canonical-pad visual reproduction

ShiboSoftwareDev:
## Problem

On current `main`, the exact final Pipeline9 input from the 96-component Allwinner T113-S3 Linux board stops at `No path found for source_trace_194`. The existing exact regression contains 29 remaining connections, 397 obstacles, and 342 preloaded fanout traces from the real four-layer board.

The next failure is a connectivity-representation mismatch: generated fixed routes identify their canonical net, while board obstacles identify member traces and ports. Regional candidate validation can therefore reject the route's own pad as foreign copper.

## Reproduction

This PR adds the pre-fix visual baseline for that existing exact regression. `t113-linux-exact-pipeline9-state.svg` renders the complete board and all 342 preloaded traces at the failure boundary. It uses the same PCB renderer, dimensions, viewport, colors, and copper layers as the fixed image.

#2581 is based on this branch and updates this exact SVG path with the same board plus the 42 routes Pipeline9 reaches after accepting the same-net candidate. GitHub's visual diff therefore shows the routing change directly.

Fixture SHA-256: `fd6a9156391178e258dc8cb4b1523e609b8587119f1a9302579b4600bba1aa63`

## Validation

- `bun test --timeout 120000 tests/bugs/t113-linux-exact-pipeline9-root.test.ts`: pass, 13 assertions
- Baseline SVG rasterized and inspected at 2400 × 1800
- `git diff --check`: pass

Stacked fix: #2581


ShiboSoftwareDev:
Superseded by #2576, which already merged this exact T113 reproduction into main. PR #2581 has been rebased directly onto current main and now carries only the canonical obstacle-ownership fix.

PR #2581, opened 2026-09-14
Resolve canonical roots in Pipeline9 via validation

ShiboSoftwareDev:
## Problem

After the closed-route serialization fix on `main`, Pipeline9 reaches a second ownership mismatch in regional via validation. Generated fixed routes carry canonical `rootConnectionName` values, while board obstacles carry member trace and port IDs. Passing those two forms directly to the generic connectivity API makes the exact T113 route `source_trace_44_fixed_262_13` treat its own `pcb_smtpad_139` as foreign copper.

The full exact run consequently stops at `No path found for source_trace_194` even though the rejected candidate only intersects copper from its own net.

## Change

- Resolve route and obstacle identities to canonical nets at the Pipeline9 regional obstacle-validation boundary.
- Map member IDs before considering an ID as a canonical net key. This preserves the collision case where one string is both a generated net key and a member of another net.
- Keep `circuit-json-to-connectivity-map@0.0.19` and its existing member-ID semantics.
- Add a focused regression that loads the merged, unmodified 96-component T113 fixture and checks the exact same-net pad plus a foreign-net obstacle.
- Update the full exact integration test to prove the solver advances through `source_trace_194` and reaches the next independent reconstruction invariant.

## Visual diff

This PR is stacked on #2592. That PR adds `t113-linux-exact-pipeline9-state.svg` with the exact board and its 342 preloaded traces at the failure boundary. This PR modifies the same file using the same renderer and viewport, adding the 42 routes reached after the canonical ownership fix. The Files changed visual view is therefore a direct before/after of this fix.

## Validation

- Exact canonical-obstacle, NRF ID-collision, and full T113 tests: 3 passed, 30 assertions
- Full exact solver run: 42 routes, then `breakout:pcb_breakout_point_68_fixed_168_1`
- `bun run build`: pass, including declarations
- Fixed SVG rasterized and inspected at 2400 × 1800
- `git diff --check`: pass

Visual reproduction: #2592

Next stacked fix: #2585


seveibar (approved):

seveibar:
@vreview

vreview:
### Video review

**Complete**

[▶ Watch the video review](https://0hmx.com/artifacts/video-review-924883209-issue_comment-5689079986-335dcc3e/)

Reviewing commit `335dcc3e`.

This comment updates as work progresses.

<sub>Job: 924883209-issue_comment-5689079986</sub>

PR #2583, opened 2026-09-15
Advance T113 routing through closed preloaded routes

ShiboSoftwareDev:
## Problem

After #2581 fixes Pipeline9's regional pad ownership, the exact 96-component T113 fixture still exposes a platform-specific closed-route serialization failure. Route 201 starts and ends at port `35301` and has two assigned segments, but the pinned `tiny-hypergraph` serializer can treat the shared endpoint as completion before consuming either segment. Linux then stops with `Route 201 could not determine endpoint regions`.

## Change

- Pin `tiny-hypergraph` to `b1c6473a04a8bc2562a3a6df5029d90873515344`, the merged root fix from tscircuit/tiny-hypergraph#188.
- Run the exact T113 fixture through port-point pathing and require every closed-route segment to serialize without the Route 201 error.
- Preserve #2581's PCB artifact. A macOS run already reaches the later 42-route state shown there; this layer makes Linux reach the same stage rather than introducing different geometry.

This PR is stacked on #2581 and changes only closed-route serialization. The next independent invariant remains `breakout:pcb_breakout_point_68_fixed_168_1`.

## Validation

- Exact T113 full-board regression through port-point pathing: pass, 15 assertions
- Exact upstream T113 Route 201 reproduction in `tiny-hypergraph`: pass
- Corrected #2581 dependency stack retained: `circuit-json-to-connectivity-map@0.0.19`

Parent fix: https://github.com/tscircuit/tscircuit-autorouter/pull/2581

Next stacked fix: https://github.com/tscircuit/tscircuit-autorouter/pull/2585


ShiboSoftwareDev:
Superseded by #2579, which already merged the focused closed-route serialization fix into main. The dependency SHA in this PR has since diverged with unrelated tiny-hypergraph changes, so it is intentionally excluded from the cleaned stack. PR #2585 is now based directly on #2581.

PR #2609, opened 2026-09-16
Detect through-hole via collisions in reference DRC

Abse2001:
On boards without blind or buried vias, a top-to-inner-layer route transition still drills through the full stack. The reference DRC now checks the via on every board layer, while `allowBlindAndBuriedVias` retains the declared span. The DRC projection does not mutate route geometry or its `from_layer`/`to_layer` path fields.

This PR leaves `high-density-repair03` at the same `5f6c9af` commit used by `main`. No Repair03 implementation, Pipeline 9 routing, via-placement, or Repair04 source change is included. This isolates the final benchmark DRC correction from route repair behavior.

Validation: all nine CI test shards, typecheck, build, formatting, and the affected visual snapshots pass. The isolated same-machine [Dataset01 benchmark](https://github.com/tscircuit/tscircuit-autorouter/pull/2609#issuecomment-5744093652) reports 85/85 completion, 85/85 DRC pass, zero timeouts, and no changed outcomes. The isolated [SRJ18 benchmark](https://github.com/tscircuit/tscircuit-autorouter/pull/2609#issuecomment-5744093716) keeps completion at 81.3% and one timeout on both sides, but the corrected final evaluator reports 49 DRC issues and four changed outcomes across samples 8, 10, 13, and 16. This confirms that the SRJ18 DRC result change exists without the Repair03 #136 implementation.


Abse2001:
@seveibar I narrowed this PR to the DRC-only behavior we discussed: route data and routing stages are unchanged, and the Repair03 pin is the previous solver commit plus its DRC projection fix ([Repair03 #137](https://github.com/tscircuit/high-density-repair03/pull/137), green CI).

The stricter DRC now exposes violations in existing solved routes. Locally, SRJ18 sample 8 reports 4 errors and sample 16 reports 2, while both integration tests currently require zero. Sample 13 also reports clearance errors in the earlier DRC-only CI run. The visual snapshot changes are being updated separately.

How would you like the zero-error integration tests handled within this DRC-only PR? I can make them assert the newly detected violations and leave routing repair for a separate follow-up, or keep them failing until a separate route fix is ready. I will not change routes or weaken DRC to satisfy them.

Abse2001:
@seveibar The full paired benchmark gives a clearer scope question. Dataset01 remains 85/85 DRC-clean. On SRJ18, completion is unchanged at 13/16, but the corrected DRC reports 37 issues: sample 8 has 4, sample 13 has 31, and sample 16 has 2. Relaxed DRC pass falls from 13/16 to 10/16. [Benchmark result](https://github.com/tscircuit/tscircuit-autorouter/pull/2609#issuecomment-5700071383).

Although this PR changes no routing source code or route `from_layer`/`to_layer` values in the DRC projection, Repair03 consumes its indexed DRC during repair. New DRC findings can therefore affect its chosen output routes (and some visual snapshots change). Is that feedback effect within the intended DRC-only scope? If exact solver outputs must remain unchanged, we may need to keep the existing Repair03 pin and apply the physical-layer rule only in final/reference DRC until the routing behavior is reviewed. I have not made that scope change.

lib/testing/utils/convertToCircuitJson.ts
```diff
@@ -800,11 +800,29 @@ function createPcbPadElements(srj: SimpleRouteJson): AnyCircuitElement[] {
  * @param minViaDiameter Default diameter for vias
  * @returns An array of PcbVia elements
  */
+function getPhysicalViaLayers(
```
seveibar, line 803: this "Physical" this is BULLSHIT. Confusing AF

```suggestion
function getViaDrillLayers(
```

seveibar: this paradigm isn't even correct, even if a via drills on a layer that doesn't mean it has a copper pad on that layer, so it's not sufficient, let's be specific so that we're at least only screwing outselves a tiny bit when we eventually want to have copper pad exclusions on layers in the middle of the span

seveibar (commented):
pretty clearly introduces a bug if i'm not mistake?

update your snapshot to include the DRC count, many snapshots have DRC count ON the visual snapshot which helps catch this

seveibar (approved):
see important comment

lib/testing/utils/convertToCircuitJson.ts
Abse2001: Updated to `getViaDrillLayers` and `drillSpan`. The helper now describes only the drilled layer extent; it does not claim that every drilled layer has a copper pad. Route `from_layer`/`to_layer` fields remain the traveled path span.

Abse2001:
Superseded by #2664, rebuilt as one clean autorouter-only commit from current main. The replacement incorporates the review feedback, keeps route output unchanged, and includes no Repair03 integration.

PR #2630, opened 2026-09-17
Improve trace widths after hypergraph pathing

Abse2001:
Pipeline 9 can choose a narrow hypergraph corridor for a trace with a wider requested width, then narrow the copper even when a wider detour exists (issue #1721, repro #2606). This PR adds a width-improvement stage after port-point pathing. It seeds each candidate with the completed assignment, penalizes ports beside regions too narrow for the requested width plus clearance, and accepts a reroute only if the estimated bottleneck improves, every other assignment stays the same, and port redistribution adds no unavoidable crossings. The initial pathing objective is unchanged; unsuccessful candidates leave its solution intact. Attempts are capped at eight per board.

The improved assignment reaches high-density routing and then trace simplification. For connections the width stage actually improved, simplification uses the requested width and pad clearance when choosing shortcuts, followed by a vertex cleanup pass. Other connections keep the existing simplification behavior. The #2606 repro now produces a clean six-point detour instead of eleven stitched points, keeps every wire at 0.4 mm, and emits no width warning.

The per-connection `minTraceWidth` survives connection merging and participates in width selection. If clearance still forces a narrower result, `TraceWidthSolver.widthWarnings` and the console warning identify the trace, requested width, and width used. A closed-corridor test keeps the completed narrow route and emits that warning. A multi-route test rejects a candidate that would cause an impossible crossing during port redistribution. Synthetic routes for preloaded copper are excluded from width improvement.

Validation: local build, TypeScript check, focused width and simplification tests, SRJ18 sample 4 regional work-budget test, and the full Game Boy Advance snapshot test. CI passes all nine test shards, format, type-check, build, and Vercel. The updated repro snapshot shows the simplified 0.4 mm detour.

Refs #1721.


seveibar (approved):

seveibar:
@Abse2001 don't merge without confirming no benchmark regressions!

Abse2001:
@seveibar The conflict is resolved on final head `a31dcc2`, and CI is green.

The paired same-machine Pipeline 9 benchmarks show no functional regressions:

- [dataset01](https://github.com/tscircuit/tscircuit-autorouter/pull/2630#issuecomment-5741948394): 85/85 complete and DRC-clean, 0 timeouts, 0 outcome changes. Total stage time is effectively unchanged (302.468s main, 302.161s PR). P95 moved from 11.0s to 12.0s while P50-P80 improved, so the aggregate result is neutral.
- [SRJ18](https://github.com/tscircuit/tscircuit-autorouter/pull/2630#issuecomment-5741948482): completion and DRC pass remain 81.3%, with 0 DRC issues, 1 timeout on both revisions, and 0 outcome changes. Total stage time improved from 1997.255s to 1977.685s.

The new width-improvement stage used 0.004s on dataset01 and 0.003s on SRJ18.

PR #2638, opened 2026-09-18
Repro: autorouter emits 0.08 mm terminal segments below 0.15 mm minimum

GokulPandi-M:
## Repro

The full autorouting input from `@tsci/gokul.acoustic-guitar-tuner@1.0.10` is captured in [autorouting bug report 106-347963](https://api.tscircuit.com/autorouting/bug_reports/view?autorouting_bug_report_id=347963bd-4459-4cad-9bea-298951ad85f5). It is a 2-layer board with 30 connections, 139 obstacles, and `minTraceWidth: 0.15`.

Running that complete SRJ through `AutoroutingPipelineSolver` on current `main` reports success, but the simplified output contains microphone-terminal wire segments around `0.0800 mm`. This is below both the SRJ minimum of `0.15 mm` and the `0.10 mm` minimum that later produces KiCad DRC errors.

The test runs the complete solver, asserts the actual undersized output, and confirms the violating geometry belongs to the `source_net_0` microphone GND connection.

The original full routed-board snapshot is preserved unchanged for context:

Full acoustic tuner autorouting repro

Screenshot 2026-09-20 at 1 22 15 PM

### Focused microphone-terminal view

The additional snapshot zooms into the affected microphone GND terminal. It keeps the original nearby routed geometry, adds the continued `0.0800 mm` red width guide, and shows the current trace width beside the required `0.1500 mm` minimum. A 3% visual tolerance covers small run-to-run route movement; the numeric width assertions remain strict.

KiCad track-width DRC errors at 0.0800 mm

## Expected

Terminal tapering may reduce the nominal `0.50 mm` GND trace to enter the microphone's polygon land, but it must not emit wire widths below the board's `0.15 mm` minimum. If the terminal geometry cannot satisfy that constraint, the router should return a clear routing/constraint failure instead of silently generating undersized copper.

This PR is repro-only and does not change solver behavior.

## Validation

- `bun test tests/bugs/bugreport106-347963.test.ts` — 1 pass, 0 failures, 6 assertions.
- `bunx tsc --noEmit` — pass.
- Biome formatting and `git diff --check` — pass.
- The full routed-board snapshot is byte-for-byte identical to the pre-zoom version.


imrishabh18 (changes requested):
Not the right way to add a repro, add the full repro circuit. Check the package.json on the script to create that

GokulPandi-M:
Updated this using the official `bug-report-with-test` flow from autorouting bug report `347963bd-4459-4cad-9bea-298951ad85f5`. I removed the reduced repro and added the full SRJ fixture, generated full-pipeline test, and full snapshot. The current pipeline emits a minimum wire width of `0.0800088 mm` with `minTraceWidth: 0.15`. The focused test passes locally. Could you take another look?

0hmX (approved):

GokulPandi-M:
merge please @0hmX @imrishabh18

seveibar (commented):
what is the bug?

seveibar:
the trace width?

GokulPandi-M:
> the trace width?

Yeah, it’s the trace width at the microphone GND terminal. The route is nominally 0.5 mm, but the autorouter tapers the terminal segment to 0.08 mm even though minTraceWidth is 0.15 mm. @seveibar

imrishabh18:
> > the trace width?
> 
> Yeah, it’s the trace width at the microphone GND terminal. The route is nominally 0.5 mm, but the autorouter tapers the terminal segment to 0.08 mm even though minTraceWidth is 0.15 mm. @seveibar

It's tapering because `0.15mm` is not possible with that layout, there is no space to increase the width of the trace without touching other pads

GokulPandi-M:
> > > the trace width?
> > 
> > 
> > Yeah, it’s the trace width at the microphone GND terminal. The route is nominally 0.5 mm, but the autorouter tapers the terminal segment to 0.08 mm even though minTraceWidth is 0.15 mm. @seveibar
> 
> It's tapering because `0.15mm` is not possible with that layout, there is no space to increase the width of the trace without touching other pads

@imrishabh18 ig , gnd pad is curved, but tscircuit splits it into small rectangular obstacles for routing. The router picks a narrow 0.08 mm section from those rectangles, even though 0.15 mm fits inside the actual pad. So it looks like a polygon width-calculation issue, not a lack of space ,imo

seveibar (changes requested):
to me, this repro doesn't look valid, what is generating that simple route json. The bug doesn't seem to be in the autorouter- it seems to be in the code generating this simple route json or maybe your TSX

why is such an odd obstacle shape created?

PR #2646, opened 2026-09-19
Connect same-region connections from degenerate HD routes

maci0:
Replaces #2561, which the stale bot closed without review, and #1530 before it.

When both endpoints of a connection sit in one capacity node, terminal ports are dropped from the HD node graph and the solver can return a single-point route. The net is then left open. Fall back to a direct segment between the connection endpoints (with a via if the layers differ).

Rebased onto current main (`8fae1e7c`); no conflicts. `tests/utils/convertHdRouteToSimplifiedRoute.test.ts` passes (9 tests, including the degenerate same-region repro), the full `tests/utils/` suite passes (47 pass / 1 skip), and `tsc --noEmit` is clean.


maci0:
Still current and CI-green; flagging to avoid the stale auto-close. Ready for review whenever someone has a moment.

PR #2642, opened 2026-09-18
Repair native high-density nodes before stitching

seveibar:
Repair actual copper-clearance violations inside every native high-density node before stitching. Pipeline4/7/8/9 pass the board's minimum trace width and connectivity into repair02, and the old node-count cutoff stays removed. The per-node inputs include immutable nearby copper and net/layer-aware pads, so repair does not depend on the order in which other nodes finish.

The previous implementation could approve moves using 0.10 mm node traces even when the board required 0.15 mm copper. It also forced DRC-clean traces away from node boundaries, failed to move free interior corners next to fixed ports, and grouped separate via contacts together. Repair02 now handles these cases directly, budgets work by conflict count, and avoids repeated conflict-key and frame-stat computation. Boundary proximity alone no longer triggers a redraw. No new repair03 calls or bounded-board reconstruction are introduced.

Validation:
- Repair02 PR https://github.com/tscircuit/high-density-repair02/pull/71: 92 tests and CI green.
- The previously regressing SRJ20 sample169, bugreport88, and Pipeline9 bugreport94 finish with zero reference DRC errors. The large boards also pass on Linux. The small BGA case now uses 34 vias versus 33.
- On bugreport94, independently checking the same native input before and after node repair reduces pre-downstream reference DRC errors from 215 to 173. The remaining three new error IDs are existing accidental contacts becoming positive-clearance violations, not new contacts.
- Regression coverage verifies all 81 nodes are processed, physical minimum trace width, fixed node ports, unchanged neighboring copper, separate via contacts, and rounded through-hole boundaries.
- All nine autorouter integration shards, type checking, formatting, and production/preview builds pass on `a7bb956`: https://github.com/tscircuit/tscircuit-autorouter/actions/runs/35412957566. Updated routing snapshots for the final revision. Sample 4 remains DRC-clean and within the unchanged region/candidate/search-node limits; its reference-validation expectation changes from 9 to 11 (main used 6).

Final dataset18 same-machine benchmark on `268a3fa` against main `8fae1e7`: https://github.com/tscircuit/tscircuit-autorouter/actions/runs/35411493269.

| Metric | Main | PR |
| --- | ---: | ---: |
| Completed / DRC-clean | 13/16 | 13/16 |
| Timeouts | 1 | 1 |
| P50 | 88.8s | 85.5s (-3.6%) |
| P90 | 241.8s | 234.8s (-2.9%) |
| Matched completed boards: global + joint DRC | 360.9s | 293.6s (-18.6%) |
[...]

PR #2647, opened 2026-09-19
Resolve different-net same-layer crossings with a via instead of truncating

maci0:
Replaces #2553/#2561-era work on #1507.

The capacity pipeline models same-layer crossings as via-resolvable *cost* (`calculateCrossingProbabilityOfFailure`), not a hard constraint. When HD/repair cannot actually resolve one, it survives into the output as a different-net same-layer overlap — a real short. Nothing downstream checked for that.

**What this does now**

`guaranteeNoSameLayerShorts` runs at the public output of Pipeline 7 (`AutoroutingPipelineSolver`) and Pipeline 9, and resolves each violation the way a human would: keep the trace's copper, run the offending segment on another layer, and transition with a via at each end.

It deliberately does **not** truncate the victim trace any more. Truncation disconnects the connection and the DRC evaluator reports a brand-new `missing_connection_*` error, so the old behaviour traded one DRC failure for another (`bugreport99`) instead of fixing anything. The safety net now only ever improves the output:

- a crossing it cannot resolve without a reroute is left as routed for the router to fix, and
- the relocation is refused when either via would land in a region with a single available layer (no layer to transition to) — the pipelines pass those regions from the component topology view, plus their board layer names, so a spare layer that is not routed yet is still usable.

**Tests**

- Unit coverage for relocating a crossing, leaving an unresolvable one, and honouring a forbidden via region; `same-layer-short-crossbar-repro` still shows DRC overlap errors going to zero.
- Local `tests/bugs/` + `tests/utils/` (CI timeout): 5 failures — `bugreport78/87/88/91/94` — which fail identically on a clean `a9bb99bd` worktree, so they are pre-existing local environment failures, not from this branch. `bugreport99`, `bugreport68`, `bugreport103/104/96`, `t113` all pass.
- CI green on `cae43760`.

```
Fixes #1507
```


maci0:
Main moved to `a9bb99bd` (#2642, "Repair native high-density nodes before stitching") while CI ran, so I rebased again:

- Conflict was the `bugreport44-0ec411` snapshot again; took main's version and regenerated (`BUN_UPDATE_SNAPSHOTS=1`), since both #2642 and this change move route geometry.
- Re-verified after the rebase: `bugreport44`, `bugreport44-source-net-0-stitch`, `same-layer-short-crossbar-repro`, `guarantee-no-same-layer-shorts`, `convertHdRouteToSimplifiedRoute` — 12 pass, 0 fail. `tsc --noEmit` clean.

The 8 `tests/bugs` failures noted earlier are still pre-existing on main (confirmed individually against a clean main worktree), not from this branch.

maci0:
Status after the latest CI run: one of the two failing shards is fixed, the other needs a maintainer call.

**Fixed — snapshot under stale deps.** My local `node_modules` was missing 52 packages relative to `package.json`, so the `bugreport44-0ec411` snapshot I regenerated did not match what CI produces. After `bun install` I regenerated it and it passes locally (590a7378).

**Improved — crossing resolution.** The safety net used to always truncate the victim trace, which produces a `missing_connection` DRC id. It now prefers the human fix: keep the copper, run the offending segment on another layer, and transition with a via at each end. Truncation remains the last resort. Unit tests and `same-layer-short-crossbar-repro` still pass.

**Blocked — `bugreport99`.** This fixture genuinely ships a different-net same-layer overlap today: at the HD output, `source_trace_52__source_trace_55_mst1@z0` and `source_trace_51__source_trace_53_mst1@z0` overlap by 0.15mm on `top` (net127 vs net123). So on `main` the short survives to the final output, and the test only requires "no **new** DRC ids", which the short satisfies because it is present in the pre-power traces too.

Enforcing "never emit a short" turns that into `missing_connection_source_trace_51__source_trace_53_mst0_0_pcb_port_1`, which the test counts as a new id. On this 2-layer fixture the overlap cannot be resolved without a reroute: moving either segment to `bottom` collides with `source_trace_13`, and cascading that relocation back to `top` re-collides with `source_trace_52`. So the choice here is ship the short (today's behavior) or ship a missing connection.

How do you want that tradeoff handled — should the guarantee leave the short in place when no alternative layer is free, or should `bugreport99` accept the missing connection as the lesser failure? Happy to take whichever direction; if the answer is "make the router not produce the overlap", that is a Pipeline 7 HD/exact-DRC change and I can open it separately.

maci0:
Correction to my last comment: I backed the via-relocation out.

It resolved crossings by moving the victim segment to another layer, but the inserted vias land wherever the crossing is — including inside the single-layer BGA regions that `bugreport68` asserts must contain no routed via. So shard 6 traded `bugreport44` for a new `bugreport68` failure. Relocation needs to respect routing regions first; that is a bigger change than a post-process safety net.

Current head (5dd2b302) is truncation-only again, with the `bugreport44` snapshot regenerated after a clean `bun install` (my earlier snapshot came from a stale `node_modules` that was missing 52 packages, which is why CI saw a 62% diff). Expected CI: shard 6 green, shard 7 red on `bugreport99` only — the short-versus-missing-connection tradeoff described above, which is the open question.

maci0:
Still current and CI-green; flagging to avoid the stale auto-close. Ready for review whenever someone has a moment.

PR #2650, opened 2026-09-19
route rv1106 across copper pours with shared free-layer regions

mohan-bee:
### Motivation
- Preserve shared layer transitions across the RV1106 copper pour while reducing fragmentation. Stacked on [the repro](https://github.com/tscircuit/tscircuit-autorouter/pull/2534) and integrating [rectdiff #148](https://github.com/tscircuit/rectdiff/pull/148).

### Before
- The base keeps top and lower-layer free regions separate. The initial overlap fix introduces many smaller regions.

### After
- Pin rectdiff to `fb1037f413e2ad32f5f6d8e9e1a7145239523259`, supply transit clearance, and merge compatible neighboring free rectangles. Refresh the RV1106, pedometer and SRJ18 sample 3 snapshots.
- All three focused tests pass locally with the pinned rectdiff source loaded against workspace dependencies; CI for the new head remains pending. Full RV1106 routing remained unfinished at the 60-second observation limit with longest-segment warnings, so final routing and DRC are not yet verified.

| Before | After |
| --- | --- |
| RV1106 pathing before | RV1106 pathing after |


0hmX:
I think I have to reject this because what this is doing is increasing  small regions and making this an grid solver.

Our ideology. Get the biggest obstacle free regions on the board because we can catch that in memory.

PR #2672, opened 2026-09-21
prevent disconnected traces when nets revisit routing regions

mohan-bee:
### Motivation
- Fix the disconnected copper reproduced in #2671. This PR is stacked on that full-board repro.

### Before
- The polyline candidate collapses repeated port pairs into one first-to-last route, leaving two isolated copper fragments.

### After
- Reject unsupported multi-pair inputs so the existing portfolio routes the requested pairs. The full-board regression verifies one continuous green-LED trace and connected region ports.
- Refresh the standard autorouter bug-report snapshot and affected board snapshots. The focused regression and formatting check pass locally.

| before | after |
| --- | --- |
| before: disconnected fragments | after: connected routing |


tests/bugs/__snapshots__/stm32-full-board-disconnected.snap.png
imrishabh18, line 1: We don't add pcb snapshots in this repo, add bug report snapshots!

imrishabh18 (changes requested):

imrishabh18 (commented):
This is just rejecting the unsupported input, rather than fixing the actual bug I guess. I am not sure maybe let's see what @seveibar has to say

PR #2673, opened 2026-09-21
Speed up trace simplification with geometry-first filtering

imrishabh18:
Trace simplification repeatedly scans long obstacle connectivity lists before checking whether those obstacles are close enough to matter. It also computes exact distances for distant peer segments. On an isolated replay of bugreport107-board-1726's simplification input (763 routes, 46,537 points, 1,100 obstacles), this change reduced runtime from 21.38 s to 3.21 s with identical complete output routes.

Changes:
- Check obstacle proximity before connectivity and remove the duplicate connectivity filter in SingleSimplifiedPathSolver5.
- Check segment/via containment before net connectivity when marking through-obstacle geometry.
- Reject peer segments outside the clearance-expanded route bounds before the existing exact-distance check. Boundary-touching candidates still receive the exact check.

Validation:
- 58 existing tests passed across simplification, via removal/merging, crossing-via reduction, immutable routes, and terminal-layer handling.
- Added and passed a regression test comparing filtered segments against exhaustive exact-distance filtering at edges, corners, and clearance boundaries, with and without trace-width-aware clearance.
- bun run build passed, including declaration generation.
- Replayed the saved bugreport107 simplification input against this branch and verified deep equality of all 763 output routes against the unchanged solver. Timings are local stage measurements, not a full-pipeline speedup or a timing assertion in CI.


seveibar (changes requested):
afaik this isn't valid. Compute box gaps is as fast as the bounds overlap. This probably causes a performance regression. You should have more detailed measurements to assess if this is a performance improvement

seveibar:
@imrishabh18 don't be tricked by random noise in performance results. You may need a dedicated simplification benchmark (in the appropriate repo, possibly refactoring simplification out of this repo which is the first step i would take) because your benchmarking is suffering from noise. We don't merge "possible" improvements, the improvement must be definitively proven

